PR #1, opened 2020-04-24
First export apprach

TheKeymaster:
* Running this will just create a CSV based on some static data.
* This is more of a documentation on how we want the structure to look like, than an actual implementation.


TheKeymaster:
Closed in favor of #5.

PR #2, opened 2020-05-06
Plenty 235 additional data

RimantasAdeoWeb:


src/Response/Collection/EntityCollection.php
```diff
@@ -107,22 +107,24 @@ private function callGetter(Entity $entity, string $property)
...
+            foreach ($value as $item) {
+                if (!$this->isCriteriaMatching($criteria, $item)) {
+                    continue;
+                }
+
+                return true;
+            }
+
+            return false;
+        } elseif ($value instanceof Entity) {
```
TheKeymaster, line 120: Nice :+1:!

src/Response/Entity/Attribute.php
```diff
@@ -0,0 +1,170 @@
...
+        $this->amazonAttribute = (string)$data['amazonAttribute'];
+        $this->fruugoAttribute = (string)$data['fruugoAttribute'];
+        $this->pixmaniaAttribute = (int)$data['pixmaniaAttribute'];
+        $this->ottoAttribute = (string)$data['ottoAttribute'];
+        $this->googleShoppingAttribute = (string)$data['googleShoppingAttribute'];
+        $this->neckermannAtEpAttribute = (int)$data['neckermannAtEpAttribute'];
+        $this->typeOfSelectionInOnlineStore = (string)$data['typeOfSelectionInOnlineStore'];
+        $this->laRedouteAttribute = (int)$data['laRedouteAttribute'];
+        $this->isGroupable = (bool)$data['isGroupable'];
+        $this->updatedAt = (string)$data['updatedAt']; //Undocumented
```
TheKeymaster, line 72: I guess we can remove this comment. This is probably automatically added by Laravel, which Plentymarkets uses internally as their backend.

src/Response/Entity/Property.php
```diff
@@ -0,0 +1,225 @@
...
+    /** @var mixed */
+    private $availabilities;
+
+    /** @var Name[] */
+    private $names = [];
+
+    /** @var Option[] */
+    private $options = [];
+
+    /** @var mixed */
```
TheKeymaster, line 51: It is probably some sort of array. I think it is better that we just add `array` as typehint.

src/Response/Entity/Property/Amazon.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /** @var string */
+    private $createdAt;
+
+    /** @var string */
+    private $updatedAt;
+
+    public function __construct(array $data)
+    {
+        //Undocumented - the properties may not match the received data exactly
```
TheKeymaster, line 34: I do not know if a code style rule for that exists, but we have made a decision a long time ago that all comments regardless of their length should always use the `//` syntax and must have a space before the comment text. E.g.

```php
class AwesomeClass
{

    public function niceFunction()
    {
        // This is a long comment. It does not exceed 120 characters. Once it would the text will be
        // moved into a new line. Additionally comments always start with a capital letter and
        // always use the `//` syntax, no matter how long the comment is.
    }
}
```

TheKeymaster: Also I would add such notices to the respective getters. This way someone that is using this class to get these contents, will know about it.

Please adapt all your comments regarding this information. This also may apply to all getters of a class (e.g. `Property`).

src/Response/Entity/Property/Selection/Relation.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /** @var RelationValue[] */
+    private $relationValues = [];
+
+    public function __construct(array $data)
+    {
+        //Undocumented - the properties may not match the received data exactly
+        $this->id = (int)$data['id'];
+        $this->propertyId = (int)$data['propertyId'];
+        $this->relationTypeIdentifier = $data['relationTypeIdentifier']; //Unknown type - received only null values
```
TheKeymaster, line 41: An identifier is typically an `int`. I am pretty sure here, as we also used this in the current plenty REST exporter. So that one would be `?int`, as a relation is not mandatory in this case.

src/Response/Entity/SalesPrice/CustomerClass.php
```diff
@@ -0,0 +1,61 @@
...
+            'updatedAt' => $this->updatedAt
+        ];
+    }
+
+    public function getSalesPriceId(): int
+    {
+        return $this->salesPriceId;
+    }
+
+    public function getCustomerClassId(): int
```
TheKeymaster, line 47: Can we change the property and this getter to just `id`/`getId`, since we are already inside of `CustomerClass`, and we do not need to repeat the context.

src/Response/Entity/VatConfiguration.php
```diff
@@ -0,0 +1,153 @@
...
+    {
+        $this->id = (int)$data['id'];
+        $this->countryId = (int)$data['countryId'];
+        $this->taxIdNumber = (string)$data['taxIdNumber'];
+        $this->startedAt = (string)$data['startedAt'];
+        $this->invalidFrom = !is_null($data['invalidFrom']) ? (string)$data['invalidFrom'] : null;
+        $this->locationId = (int)$data['locationId'];
+        $this->marginScheme = (string)$data['marginScheme'];
+        $this->isRestrictedToDigitalItems = (bool)$data['isRestrictedToDigitalItems'];
+        $this->isStandard = (bool)($data['isStandard'] ?? false); //documentated as bool, however nothing was returned during testing
```
TheKeymaster, line 57: At this point I just want to thank Plentymarkets that they add documentation for things that do not exist but also thank them for not providing any documentation for things that do exist. We have so many classes where Plenty just does not provide any documentation. How sad.

tests/Response/Collection/ItemPropertiesResponseTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+class ItemPropertiesResponseTest extends TestCase
+{
+    use ResponseHelper;
+
+    private $response;
+
+    private $propertiesResponse;
+
+    public function setup(): void
```
TheKeymaster, line 21: The method is actually called `setUp`. I have seen you have used `setup` in multiple places. Please change it in all tests to match `setUp`.

src/Response/Entity/SalesPrice/CustomerClass.php
RimantasAdeoWeb: That is literally the name this property is returned as though. Are you sure about this?

src/Response/Entity/Property/Selection/Relation.php
RimantasAdeoWeb: I'm assuming it's the same for relationTargetId, right? Casting that to int as well.

src/Response/Entity/SalesPrice/CustomerClass.php
TheKeymaster: Yes I also did this for the category. It is just easier to read and as an implementer you use your IDE to get the things that you need anyway. If you want to know where the data comes from at any point, you can always go ahead and check the implementation.

src/Response/Entity/Property/Selection/Relation.php
TheKeymaster: Yes makes sense.

PR #5, opened 2020-06-16
Plenty 240 csv wrapper

RimantasAdeoWeb:


src/RegistryWarmer.php
```diff
@@ -0,0 +1,243 @@
...
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\SalesPriceResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\AttributeResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ManufacturerResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ItemPropertyResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\UnitResponse;
+use GuzzleHttp\Client as GuzzleClient;
+use InvalidArgumentException;
+use Psr\Log\LoggerInterface;
+
+class RegistryWarmer
```
TheKeymaster, line 42: `RegistryWarmer` => `RegistryService`. This way we are not limited to only warming up the registry. Maybe we want to add some helper methods or something here in the future.

src/RegistryWarmer.php
```diff
@@ -0,0 +1,243 @@
...
+        ?Registry $registry = null
+    ) {
+        $this->internalLogger = $internalLogger;
+        $this->customerLogger = $customerLogger;
+        $this->config = $config;
+        $this->client = $client ?? new Client(new GuzzleClient(), $config, $internalLogger, $customerLogger);
+        $this->registry = $registry ?? new Registry();
+    }
+
+    public function warmUpRegistry(): void
```
TheKeymaster, line 73: It is clear that we are in the context of this class, which has something to do with the registry. Therefore, you can remove the `...Registry` part and just leave it as `warmUp`.

src/Wrapper/Product.php
```diff
@@ -0,0 +1,156 @@
...
+    protected function setTexts(): void
+    {
+        /** @var WebStore $webStore */
+        $webStore = $this->registry->get('webStore');
+
+        if (!$webStore || !$webStore->getConfiguration()) {
+            return;
+        }
+
+        $textGetter = 'getName' . $webStore->getConfiguration()['displayItemName'];
```
TheKeymaster, line 90: I tried running the exporter and it worked pretty good. Here in this case very old shops do not have a `displayItemName`, but we will fix that in a future story.

src/Wrapper/Product.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        return $this->reason;
+    }
+
+    protected function setTexts(): void
+    {
+        /** @var WebStore $webStore */
+        $webStore = $this->registry->get('webStore');
+
+        if (!$webStore || !$webStore->getConfiguration()) {
```
TheKeymaster, line 86: As we discussed this in our call, this should be only done once, before the export is even started (before starting the export of products). The registry won't change during exporting time.

src/Exporter/Exporter.php
```diff
@@ -102,227 +116,94 @@ public static function buildInstance(
...
     }
 
-    private function getItems(): ItemResponse
+    private function getItems($page): ItemResponse
     {
-        $itemRequest = new ItemRequest(null, $this->config->getLanguage());
-
-        $items = [];
+        $this->itemRequest->setPage($page);
+        $response = $this->client->send($this->itemRequest);
```
TheKeymaster, line 193: I have noticed that the REST exporter is logging in twice, which means that there are probably two instances of a `Client`, which shouldn't be necessary. Please check this.

RimantasAdeoWeb: Good catch, there indeed were.

PR #8, opened 2020-07-07
Plenty 250 variations

RimantasAdeoWeb:


src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+    private $id;
+
+    /** @var int */
+    private $itemId;
+
+    /** @var string[] */
+    private $barcodes = [];
+
+    /** @var array */
+    private $categories = [];
```
TheKeymaster, line 54: This should be an array of `\FINDOLOGIC\Export\Data\Attribute`.

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+                continue;
+            }
+
+            foreach ($category->getDetails() as $categoryDetail) {
+                if (strtoupper($categoryDetail->getLang()) !== strtoupper($this->config->getLanguage())) {
+                    continue;
+                }
+
+                $this->categories[] = [
+                    'categoryId' => $variationCategory->getCategoryId(),
```
TheKeymaster, line 123: We do not need the category id.

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+                    'name' => $categoryDetail->getName(),
```
TheKeymaster, line 124: The category should be exported as a libflexport `Attribute` with the key `cat`.

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+                    'url' => $categoryDetail->getPreviewUrl()
```
TheKeymaster, line 125: The url here should be set as a libflexport `Attribute` with the key `cat_url`.

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+                ];
+            }
+        }
+    }
+
+    private function processPrices(): void
+    {
+        foreach ($this->variationEntity->getVariationSalesPrices() as $variationSalesPrice) {
+            $this->prices[] = [
+                'price_id' => $variationSalesPrice->getSalesPriceId(),
```
TheKeymaster, line 135: This should be exported as `Property` with the name `price_id`.

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+                'price' => $variationSalesPrice->getPrice()
```
TheKeymaster, line 136: You may want to set this as a libflexport `Price`, or a `float`.

There is

* `Item::addPrice` Requires a `float`.
* `Item::setPrice`. Requires a `Price` object. 

You can choose whatever you prefer here.

src/Wrapper/Variation.php
TheKeymaster: Also see my other comments. It may make sense to have `$attributes` and `$properties`, which hold all sorts of attributes and properties like that:

```php
private $attributes = [];

// . . .

someMethod()
{
    $this->attributes[] = new Attribute('cat', ['whatever']);
}
```

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+            // @codeCoverageIgnoreEnd
+        }
+
+        foreach ($this->variationEntity->getVariationAttributeValues() as $variationAttributeValue) {
+            $attribute = $attributes->findOne(['id' => $variationAttributeValue->getAttributeId()]);
+            if (!$attribute) {
+                continue;
+            }
+
+            $this->attributes[] = [
```
TheKeymaster, line 158: Add this like that:

```php
$this->attributes[] = new Attribute(
    $variationAttributeValue->getAttribute()->getBackendName(),
    [$variationAttributeValue->getAttributeValue()->getBackendName()]
);
```

tests/Exporter/ExporterTest.php
```diff
@@ -311,4 +321,45 @@ private function getExpectedCsvItems(): array
 
         return [$csvItem1, $csvItem2, $csvItem3, $csvItem4, $csvItem5, $csvItem6];
     }
+
+    private function setupRegistryMockReturns(WebStore $webStore, int $variationsCount)
```
TheKeymaster, line 325: `setupRegistryMockReturns` => `setupRegistryMock`

TheKeymaster: Missing return type hint.

tests/Exporter/ExporterTest.php
```diff
@@ -311,4 +321,45 @@ private function getExpectedCsvItems(): array
...
+            $properties,
+            $units,
+        ];
+
+        for ($i = 0; $i < $variationsCount; $i++) {
+            $returnsQueue[] = $categories;
+            $returnsQueue[] = $attributes;
+        }
+
+        call_user_func_array(
```
TheKeymaster, line 360: Why are you not calling it directly? Are there any issues when doing so?

tests/Wrapper/VariationTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->exporterMock = $this->getMockBuilder(CsvFileExporter::class)
+            ->setMethodsExcept(['createItem'])
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $this->defaultConfig = $this->getDefaultConfig();
+        $this->defaultConfig->setLanguage('en');
+    }
+
+    public function testThing(): void // TODO: rename
```
TheKeymaster, line 57: Umm.. yeah :sweat_smile:

src/Wrapper/Variation.php
```diff
@@ -0,0 +1,219 @@
...
+use FINDOLOGIC\PlentyMarketsRestExporter\Config;
+use FINDOLOGIC\PlentyMarketsRestExporter\Registry;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\AttributeResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\CategoryResponse;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Entity\ItemVariation;
+
+class Variation
+{
+    /** @var Item */
+    private $item;
```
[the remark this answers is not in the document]
RimantasAdeoWeb: I will also remove the exporter from the constructor argument as the only thing it did was create the item.

tests/Exporter/ExporterTest.php
RimantasAdeoWeb: Just so I can pass the arguments as an array.

RimantasAdeoWeb: Just so I can pass the arguments as an array.

src/Wrapper/Variation.php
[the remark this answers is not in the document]
TheKeymaster: :+1:

tests/Exporter/ExporterTest.php
TheKeymaster: Ah I see. Ok then :).

src/Wrapper/Variation.php
RimantasAdeoWeb: I don't really understand the how and why of this - the Item is not accessible in the variations wrapper.

src/Wrapper/Variation.php
RimantasAdeoWeb: Is there a need for the separate categories array then? Do I just do it like this then:
```
$this->attributes[] = new Attribute('cat', [$categoryDetail->getName()]);
$this->attributes[] = new Attribute('cat_url', [$categoryDetail->getPreviewUrl()]);
```
?

TheKeymaster: Yes that was actually the plan.

src/Wrapper/Variation.php
TheKeymaster: Yes but in the Product wrapper we need to get this information and set it. Here we are already adding it but you can choose weither to do that with a `float` or the `Price` object from libflexport. Both can be consumed by libflexport at a later point.

src/Wrapper/Variation.php
RimantasAdeoWeb: Hm, is it even necessary at all? All it is is a number and if we separate it from the price, is there really a point in indexing it for search?

TheKeymaster: Properties just give additional information about the product. Since the current exporter also exported it, we will also export it. When we are done with the entire rewrite, we want to have the same export file as the current one.

RimantasAdeoWeb:
```
private function processPrices(): void
    {
        $priceIdProperty = new Property('price_id');
        foreach ($this->variationEntity->getVariationSalesPrices() as $variationSalesPrice) {
            $priceIdProperty->addValue((string)$variationSalesPrice->getSalesPriceId());
            $this->prices[] = $variationSalesPrice->getPrice();
        }

        $this->properties[] = $priceIdProperty;
    }
```

In the second foreach iteration it throws an error on addValue() saying 'Property "price_id" already has a value for usergroup "".'
Do I use the priceId itself as a fake usergroup just so I can add the value? Seems very hacky.

TheKeymaster: As I wrote on YouTrack, please use the price_id of the main variant.

RimantasAdeoWeb: That's not what I was asking about. A single variation can have multiple price_ids and I can't add multiple values to the price_id property without specifying different usergroups to go with the values. Do I use the price_ids themselves as fake usergroups or concatenate all the price_ids before setting them on the property? Or is there a difference at all?

TheKeymaster: A usergroup is not here to allow the same values for the same key e.g. `price_id`. In fact for CSV always "no usergroup" will be used `\FINDOLOGIC\Export\CSV\CSVItem::buildProperties`, so you can add them with additional usergroups, but only the "no usergroup" property will show up in the generated CSV file.

Even if the main variation has multiple `price_ids`, we will just use the first one for now. We may have to change that in a future story, but currently this data is not really important for most customers anyway.

src/Response/Entity/WebStore/Configuration.php
```diff
@@ -0,0 +1,1274 @@
...
+    public function getFeedbackCustomerNameVisibility(): ?int
+    {
+        return $this->feedbackCustomerNameVisibility;
+    }
+
+    public function getUrlTrailingSlash(): ?int
+    {
+        return $this->urlTrailingSlash;
+    }
+}
```
TheKeymaster, line 1274: Such a big class, only containing data. Thanks plenty for not splitting that stuff up...

tests/Wrapper/VariationTest.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $wrapper = new VariationWrapper(
+            $this->defaultConfig,
+            $this->registryMock,
+            $variationEntity
+        );
+
+        $wrapper->processData();
+
+        $this->assertEquals(true, $wrapper->isMain());
```
TheKeymaster, line 62: `$this->assertEquals(true, ...` => `$this->assertTrue(...)`

src/Response/Entity/WebStore/Configuration.php
```diff
@@ -0,0 +1,1274 @@
...
+        $this->attributeVariantCheck = $this->getIntProperty('attributeVariantCheck', $data);
+        $this->attributeSelectDefaultOption = $this->getIntProperty('attributeSelectDefaultOption', $data);
+        $this->urlItemCategory = $this->getStringProperty('urlItemCategory', $data);
+        $this->urlItemContent = $this->getStringProperty('urlItemContent', $data);
+        $this->urlTitleItemName = $this->getStringProperty('urlTitleItemName', $data);
+        $this->urlTitleItemContent = $this->getStringProperty('urlTitleItemContent', $data);
+        $this->urlNeedle = $this->getStringProperty('urlNeedle', $data);
+        $this->urlFileExtension = $this->getStringProperty('urlFileExtension', $data);
+        $this->urlLinking = $this->getStringProperty('urlLinking', $data);
+        $this->languageList = $this->getStringProperty('languageList', $data);
```
TheKeymaster, line 421: Please explode here directly. Anyone getting the language list, is probably converting it to an array at some point anyway, as without it it'd be pretty much useless.

tests/Response/Collection/WebStoreResponseTest.php
```diff
@@ -48,7 +48,7 @@ private function getDefaultWebStoreConfiguration(): array
             'urlNeedle' => '-',
             'urlFileExtension' => '/',
             'urlLinking' => 'absolute',
-            'languageList' => 'de, en',
+            'languageList' => ['de', 'en'],
```
TheKeymaster, line 51: Why was this changed? The response from the API shouldn't have changed and the parser should still be able to work with this format.

RimantasAdeoWeb: This method here is used in some test as the expected result of the parser, that's why I changed it here. The parser works fine with the old one as well and there are tests that cover that case.

PR #10, opened 2020-08-31
PLENTY-251 export variations

RimantasAdeoWeb:


src/Wrapper/Product.php
```diff
@@ -118,7 +119,21 @@ protected function processVariations(): void
         foreach ($this->variationEntities as $variationEntity) {
             $variation = new Variation($this->config, $this->registry, $variationEntity);
             $variation->processData();
-            $this->variations[] = $variation;
+
+            if ($variation->isMain()) {
+                $this->item->addImage($variation->getImage());
```
TheKeymaster, line 124: In case a variation does not have an image, an exception is thrown. Please add a test for that.

```
Return value of FINDOLOGIC\PlentyMarketsRestExporter\Wrapper\Variation::getImage() must be an instance of FINDOLOGIC\Export\Data\Image, null returned 
#0 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Wrapper/Product.php(135): FINDOLOGIC\PlentyMarketsRestExporter\Wrapper\Variation->getImage()
#1 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Wrapper/Product.php(72): FINDOLOGIC\PlentyMarketsRestExporter\Wrapper\Product->processVariations()
#2 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Wrapper/CsvWrapper.php(63): FINDOLOGIC\PlentyMarketsRestExporter\Wrapper\Product->processProductData()
#3 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Exporter/CsvExporter.php(66): FINDOLOGIC\PlentyMarketsRestExporter\Wrapper\CsvWrapper->wrap(0, 36, Object(FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ItemResponse), Object(FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ItemVariationResponse))
#4 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Exporter/Exporter.php(172): FINDOLOGIC\PlentyMarketsRestExporter\Exporter\CsvExporter->wrapData(36, Object(FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ItemResponse), Object(FINDOLOGIC\PlentyMarketsRestExporter\Response\Collection\ItemVariationResponse))
#5 /home/dom/github/findologic/plentymarkets-rest-exporter-new/src/Exporter/Exporter.php(163): FINDOLOGIC\PlentyMarketsRestExporter\Exporter\Exporter->exportProducts()
#6 /home/dom/github/findologic/plentymarkets-rest-exporter-new/bin/export(62): FINDOLOGIC\PlentyMarketsRestExporter\Exporter\Exporter->export()
```

src/Wrapper/Product.php
```diff
@@ -72,10 +69,10 @@ public function __construct(
     public function processProductData(): ?Item
     {
         $this->processVariations();
+
         $this->setTexts();
 
         $this->item->addSalesFrequency(0);
-        $this->item->addPrice(13.37);
```
TheKeymaster, line 78: Ensure that the item always has a price. In case it does not have a price, skip it.

src/Wrapper/Variation.php
```diff
@@ -64,18 +98,124 @@ public function __construct(
...
+
+    /**
+     * @return string[]
+     */
+    public function getBarcodes(): array
+    {
+        return $this->barcodes;
+    }
+
+    public function getPrice(): float
```
TheKeymaster, line 166: This may return `null` if the price is `0`

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+                $this->attributes[] = new Attribute($propertyName, [$value]);
+            }
+        }
     }
 
-    /**
-     * @return float[]
-     */
-    public function getPrices(): array
+    private function processVariationSpecificProperties(): void
```
TheKeymaster, line 419: These methods were probably for the most part copies from the old implementation. Please try to restructure that a bit, since that code isn't really that well structured, when I look at it.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+                    // @codeCoverageIgnoreEnd
+            }
+
+            if ($propertyName == null || $value == "null" || $value == null || $value == '') {
+                // @codeCoverageIgnoreStart
+                continue; // Ignore empty properties.
+                // @codeCoverageIgnoreEnd
+            }
+
+            if (is_array($value)) {
```
TheKeymaster, line 486: This check is redundant. libflexport can handle multiple values for a single attribute anyway.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
     }
 
-    /**
-     * @return Attribute[]
-     */
-    public function getAttributes(): array
+    private function getPropertyGroupForPropertyName(int $propertyGroupId): string
     {
-        return $this->attributes;
+        if (!$this->propertyGroups) {
```
TheKeymaster, line 498: That check should only be done once and not for each and every variant.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+            // @codeCoverageIgnoreStart
+            return '';
+            // @codeCoverageIgnoreEnd
+        }
+
+        /** @var PropertyGroup[] $properties */
+        if (!$propertyGroup = $this->propertyGroups->findOne(['propertyGroupId' => $propertyGroupId])) {
+            // @codeCoverageIgnoreStart
+            return '';
```
TheKeymaster, line 507: Make the method return value nullable and just return `null` in this case. Also add a test.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+        }
+
+        return $propertyGroup->getBackendName();
     }
 
-    /**
-     * @return Property[]
-     */
-    public function getProperties(): array
+    private function getPropertyValue(VariationProperty $variationProperty)
```
TheKeymaster, line 520: Missing return type. I know that there should be several return values, in that case add them all to the docblock.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+                // @codeCoverageIgnoreEnd
+            case 'int':
+                $value = $variationProperty->getValueInt();
+                break;
+            case 'float':
+                $value = $variationProperty->getValueFloat();
+                break;
+            default:
+                // @codeCoverageIgnoreStart
+                $value = '';
```
TheKeymaster, line 559: I prefer returning `null` here.

src/Wrapper/Variation.php
```diff
@@ -122,15 +262,37 @@ private function processCategories(): void
...
+                // @codeCoverageIgnoreStart
+                continue;
+                // @codeCoverageIgnoreEnd
+            }
+            if ($variationSalesPrice->getSalesPriceId() == $priceId) {
+                if (!$this->price) {
+                    $this->price = $price;
+                // @codeCoverageIgnoreStart
+                // not sure if this can ever actually happen
+                } elseif ($this->price > $price) {
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Yeah, recreated this piece of logic from the current exporter here, just in case.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
-    public function getBarcodes(): array
+    private function processVariationProperties(): void
     {
-        return $this->barcodes;
+        $variationProperties = $this->variationEntity->getVariationProperties();
+
+        foreach ($variationProperties as $variationProperty) {
+            $itemProperty = $variationProperty->getProperty();
+
+            if ($itemProperty && !$itemProperty->isSearchable()) {
```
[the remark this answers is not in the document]
RimantasAdeoWeb: That would make the logic differ from the current exporter though. Are you sure about this?

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
 
-    /**
-     * @return float[]
-     */
-    public function getPrices(): array
+    private function processVariationSpecificProperties(): void
     {
-        return $this->prices;
+        /** @var PropertyResponse $properties */
+        if (!$properties = $this->registry->get('properties')) {
```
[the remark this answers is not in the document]
RimantasAdeoWeb: It is checked in the current exporter in pretty much the same way as here.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
-        return $this->id;
+            foreach ($tag->getTag()->getNames() as $tagName) {
+                if ($tagName->getTagLang() === strtolower($this->config->getLanguage())) {
+                    $correctTagName = $tagName->getTagName();
+
+                    break;
+                }
+            }
+
+            $this->tags[] = new Keyword($correctTagName);
```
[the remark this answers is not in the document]
TheKeymaster: I see, maybe you want to add this detail as a docblock.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
     {
-        return $this->properties;
+        $propertyType = $variationProperty->getProperty()->getValueType();
+
+        $value = '';
+
+        switch ($propertyType) {
+            case 'empty':
+                $value = $variationProperty->getProperty()->getBackendName();
+                // break omitted intentionally.
```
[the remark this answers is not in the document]
RimantasAdeoWeb: It has to go into the next case. I don't know why, that's how the logic works in the old exporter, I just refactored it here. The comment is here just to assure the next developer that there's not supposed to be a break here because it kind of looks like there should be one.

src/Wrapper/Variation.php
RimantasAdeoWeb: The return type here is actually significant here.
`processVariationProperties()` has an `if` condition where it checks specifically for an empty string. If I change that condition to check for null values instead, other values also pass that condition which previously wouldn't - such as when a property type is float but `getValueFloat()` returns null.
This may cause a difference with the current exporter logic.

src/Wrapper/Variation.php
```diff
@@ -157,70 +319,248 @@ private function processAttributes(): void
...
+            case 'text':
+                // @codeCoverageIgnoreStart
+                // TODO: test when names structure becomes known
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Can you send me some raw response body where I could see that?

src/Wrapper/Variation.php
RimantasAdeoWeb: Can you write the exact code I should have here then? I tried figuring it out myself but then the final output differs so I'm not sure if I'm doing things correctly.

src/Response/Collection/PropertySelectionResponse.php
```diff
@@ -71,4 +71,24 @@ public function find(array $criteria): array
     {
         return $this->findEntitiesByCriteria($this->selections, $criteria);
     }
+
+    public function getPropertySelectionValues(int $id, string $lang): array
+    {
+        $propertySelections = $this->find(['propertyId' => $id]);
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Not really, since it returns propertySelections, which have at least a single relationValue with the right language, not the filtered relationValues themselves. So afterwards you need to check the languages anyway.

src/Wrapper/Variation.php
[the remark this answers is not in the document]
RimantasAdeoWeb: Seems really unnecessary. Wouldn't even know how to formulate that.

src/Wrapper/Variation.php
TheKeymaster: It shouldn't differ if you do:

```php
$this->attributes[] = new Attribute($propertyName, (array)$value);
```

src/Wrapper/Variation.php
TheKeymaster: Before starting to export products. In case it is not set throw an exception, since this information is crucial.

src/Wrapper/Variation.php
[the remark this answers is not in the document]
TheKeymaster: I am sure. That check does not make any sense. In case `$itemProperty` is `null`, we shouldn't call any methods on it.

TheKeymaster (approved):
LGTM!

TheKeymaster:
Relevant changes are already inside `develop`, therefore closing this PR.

PR #13, opened 2020-10-23
PLENTY-271-finalization

RimantasAdeoWeb:


src/Wrapper/Product.php
```diff
@@ -156,29 +163,53 @@ protected function processVariations(): int
...
-                $this->item->addPrice($variation->getPrice());
-                $this->item->setInsteadPrice($variation->getInsteadPrice());
-                foreach ($variation->getGroups() as $group) {
-                    $this->item->addUsergroup($group);
-                }
-                $this->item->setAllKeywords($variation->getTags());
             }
 
-            $this->addOrdernumbers($variation);
+            $orderNumbers = array_merge($orderNumbers, $this->getVariationOrdernumbers($variation));
```
TheKeymaster, line 187: Please write `ordernumber` as a single word, since we should try to always use the same wording.

src/Wrapper/Product.php
```diff
@@ -283,18 +314,21 @@ private function isLanguageAvailable(): bool
...
-    private function addOrdernumbers(Variation $variation): void
+    private function getVariationOrdernumbers(Variation $variation): array
     {
         $orderNumberGetters = ['number', 'model', 'id', 'itemId'];
+        $orderNumbers = [];
 
         foreach ($orderNumberGetters as $field) {
             $getter = 'get' . ucfirst($field);
-            $this->addOrdernumber((string)$variation->{$getter}());
+            $orderNumbers[] = (string)$variation->{$getter}();
```
TheKeymaster, line 324: See my other comment regarding `ordernumber`.

src/Wrapper/Product.php
```diff
@@ -156,29 +163,53 @@ protected function processVariations(): int
...
 
+        if ($prices) {
+            $this->item->addPrice(min($prices));
+        }
+
+        if ($insteadPrices) {
+            $this->item->setInsteadPrice(min($insteadPrices));
+        }
+
+        $orderNumbers = array_unique($orderNumbers);
```
TheKeymaster, line 207: Good idea to make them unique here.

tests/Wrapper/ProductTest.php
```diff
@@ -258,6 +266,9 @@ public function testProductIsSuccessfullyWrapped(): void
             'https://plenty-testshop.de/' . $expectedUrlPath . '/a-0',
             $item->getUrl()->getValues()['']
         );
+        $this->assertTrue(
+            \DateTime::createFromFormat(\DateTime::ISO8601, $item->getDateAdded()->getValues()['']) !== false
```
TheKeymaster, line 270: Please import this.

tests/Wrapper/ProductTest.php
```diff
@@ -325,6 +336,311 @@ public function testLanguagePrefixIsAddedToUrlInCaseLanguageIsAvailableButNotDef
...
+        $this->assertEquals($item->getSort()->getValues(), ['' => 2]);
+    }
+
+    public function testKeywordsAreSetFromAllVariations(): void
+    {
+        $this->exporterMock = Exporter::create(Exporter::TYPE_CSV);
+
+        $variationResponse = $this->getResponseAsArray('Pim/Variations/response.json');
+        $entries = array_slice($variationResponse['entries'], 0, 2);
+        $entries[0]['tags'] = [
```
TheKeymaster, line 373: Is the index really necessary here?

tests/Wrapper/ProductTest.php
```diff
@@ -325,6 +336,311 @@ public function testLanguagePrefixIsAddedToUrlInCaseLanguageIsAvailableButNotDef
...
+        $entries[0]['salesPrices'][1]['price'] = 150;
+        $entries[1]['salesPrices'][0]['salesPriceId'] = 0;
+        $entries[1]['salesPrices'][0]['price'] = 50;
+        $entries[1]['salesPrices'][1]['salesPriceId'] = 1;
+        $entries[1]['salesPrices'][1]['price'] = 100;
+        $entries[2]['salesPrices'][0]['salesPriceId'] = 0;
+        $entries[2]['salesPrices'][0]['price'] = 250;
+        $entries[2]['salesPrices'][1]['salesPriceId'] = 1;
+        $entries[2]['salesPrices'][1]['price'] = 300;
+        $variationResponse['entries'] = $entries;
```
TheKeymaster, line 493: This bloat distracts a bit from the actual test. Can't this be set in a separate response directly? This way the test data and the actual test is separated.

TheKeymaster: Please do this for all tests.

tests/Wrapper/VariationTest.php
```diff
@@ -188,4 +189,49 @@ public function testCharacteristicsNotAvailableForSearchAreIgnored(ItemProperty
...
+            $variationEntity
+        );
+
+        $standardVatResponse = $this->getMockResponse('VatResponse/standard_vat.json');
+        $standardVat = VatParser::parseSingleEntityResponse($standardVatResponse);
+        $this->registryServiceMock->expects($this->any())->method('getStandardVat')->willReturn($standardVat);
+
+        $wrapper->processData();
+
+        $this->expectException(\TypeError::class);
```
TheKeymaster, line 234: Please import this.

TheKeymaster: When I expect an exception I typically add it to the top of the test. This way the next developer immediately sees that I expect an exception here.

TheKeymaster: Would this exception cause an export to fail? So if my shop had a vat id of `10`, would this break the export?

TheKeymaster (changes requested):

tests/Wrapper/VariationTest.php
RimantasAdeoWeb: Yes, it would cause it to fail completely.
I also noticed that the old exporter actually takes the vat rate from the LAST variation of an item, not the first. I'll update this exporter to match the old one in this regard.
Another problem is that the old exporter would export nothing in the vat rate column in a case like this here. The libflexport exports an unset taxrate as 0. Is this difference ok?

TheKeymaster: > Yes, it would cause it to fail completely.

This is something we should avoid. I have already seen that you have changed it, but in the future we should make sure that the exporter only fails in a managed manner, and not due to some cryptic type error.

Only developers know what's wrong when checking the logs, but also others should read and understand the logs without the need of manually checking the code (if they even have dev experience).

> When changing this, I've noticed a few things:
I also noticed that the old exporter actually takes the vat rate from the LAST variation of an item, not the first. I'll update this exporter to match the old one in this regard.

That's a good call, since we want the new exporter to return the same data as the old one.

> Another problem is that the old exporter would export nothing in the vat rate column in a case like this here. The libflexport exports an unset taxrate as 0. Is this difference ok?

Yes this is fine, as libflexport should handle this right. If it turns out that this may cause issues at customers, we may need a change in libflexport, but for now this shouldn't be a problem.

TheKeymaster (approved):

PR #15, opened 2020-11-11
PLENTY-253 update url format

RimantasAdeoWeb:


tests/Wrapper/ProductTest.php
```diff
@@ -264,7 +264,7 @@ public function testProductIsSuccessfullyWrapped(): void
         $this->assertSame($expectedSummary, $item->getSummary()->getValues()['']);
         $this->assertSame($expectedDescription, $item->getDescription()->getValues()['']);
         $this->assertSame(
-            'https://plenty-testshop.de/' . $expectedUrlPath . '/a-0',
+            'https://plenty-testshop.de/' . $expectedUrlPath . '_0_0',
```
TheKeymaster, line 267: Please change the test data to something that makes this safer. Right now it wouldn't matter if id and main variation id got mixed up, since they're the same in the test.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #17, opened 2020-12-03
PLENTY-283 export variation_id, price_id, base_unit, package_size

RimantasAdeoWeb:


src/Wrapper/Product.php
```diff
@@ -177,6 +189,14 @@ protected function processVariations(): int
                 $this->item->addKeyword($tag);
             }
 
+            if ($packageSize === null) {
```
TheKeymaster, line 192: The method `processVariations` is getting big. Can we extract everything that happens only for the first variation?

TheKeymaster, line 192: Why are we being explicit about it being `null`, while on line `239` we only check for falseness? I prefer not to explicitly compare against `null`, since we're working with strict types anyway and this is either an instance or null.

src/Wrapper/Variation.php
```diff
@@ -263,12 +280,9 @@ private function buildCategoryPath(Category $category): string
 
     private function processPrices(): void
     {
-        $priceId = $this->registryService->getPriceId();
         $insteadPriceId = $this->registryService->getRrpId();
 
-        $priceIdProperty = new Property('price_id');
-        $priceIdProperty->addValue((string)$priceId);
-        $this->properties[] = $priceIdProperty;
+        $priceId = $this->registryService->getPriceId();
```
TheKeymaster, line 285: :ok_hand:

src/Wrapper/Variation.php
```diff
@@ -373,4 +387,19 @@ private function processVatRate(): void
...
+
+    private function processUnits(): void
+    {
+        $unitData = $this->variationEntity->getUnit();
+
+        $this->packageSize = $unitData->getContent();
+
+        if ($unitEntity = $this->registryService->getUnit($unitData->getUnitId())) {
+            foreach ($unitEntity->getNames() as $name) {
+                if (strtolower($name->getLang()) == (strtolower($this->config->getLanguage()))) {
```
TheKeymaster, line 399: We are working with two strings here, so we can be strict with `===`.

tests/MockData/Pim/Variations/response_for_image_test.json
```diff
@@ -75,7 +75,11 @@
       "comments": null,
       "id": 1004,
       "timestamps": null,
-      "unit": null,
+      "unit": {
```
TheKeymaster, line 78: Good idea to add this information to pre-existing responses.

tests/Wrapper/ProductTest.php
```diff
@@ -270,6 +284,15 @@ public function testProductIsSuccessfullyWrapped(): void
...
             $item->getUrl()->getValues()['']
         );
+
+        $line = $item->getCsvFragment(self::AVAILABLE_PROPERTIES);
+        $line = trim($line, "\n");
+        $columnValues = explode("\t", $line);
+        $this->assertSame((string)$expectedPriceId, $columnValues[18]);
+        $this->assertSame((string)$expectedMainVariationId, $columnValues[19]);
+        $this->assertSame('Stück', $columnValues[20]);
+        $this->assertSame('1000', $columnValues[21]);
```
TheKeymaster, line 294: Move these two strings into a separate variable and call them `$expected...`. This way it is clear what you're expecting. It is kinda hard to know what exactly `$columValues[20]` is, without having to manually check the constant.

tests/Wrapper/ProductTest.php
```diff
@@ -575,4 +598,9 @@ private function getProduct(): Product
             $this->variationEntityMocks
         );
     }
+
+    private function getExporter(): Exporter
```
TheKeymaster, line 602: Good idea to create a separate method.

TheKeymaster (changes requested):

src/Wrapper/Product.php
RimantasAdeoWeb: It's to check whether it has been set yet. I'll change the one on line 239 to match

RimantasAdeoWeb: Not really. Most of the things in the foreach are not happening for the first variation, but for the first variation with a valid value. So the the code inside of the loop sets and depends on variables outside of the loop. It's indeed getting big but refactoring it would probably make it even messier and harder to understand.

TheKeymaster (approved):

PR #19, opened 2020-12-14
PLENTY-284 export tags only for the selected store

RimantasAdeoWeb:


src/Wrapper/Variation.php
```diff
@@ -360,6 +366,21 @@ private function processTags(): void
         }
     }
 
+    private function canProcessTag(Tag $tag, int $storeId): bool
```
TheKeymaster, line 369: I prefer the naming `should`, because it theoretically always "can" be exported.

src/Wrapper/Variation.php
```diff
@@ -360,6 +366,21 @@ private function processTags(): void
...
+    {
+        if (!$clients = $tag->getTagData()->getClients()) {
+            return false;
+        };
```
TheKeymaster, line 373: What is this `;` for? Probably added by accident.

TheKeymaster (changes requested):

PR #23, opened 2021-04-01
PLENTY-248 upgrade PHP version

RimantasAdeoWeb:


composer.json
```diff
@@ -21,16 +21,16 @@
...
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.0",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/var-dumper": "^5.1",
         "symfony/console": "^5.1",
         "symfony/filesystem": "^5.1",
-        "symfony/finder": "^5.1"
+        "symfony/finder": "^5.1",
+        "phpunit/phpunit": "^9.5"
```
TheKeymaster, line 29: This looks like it has been manually edited, or somehow updated, without also updating the `composer.lock` file. 
Please run `composer require --dev phpunit/phpunit:^9.5 --with-all-dependencies` to update PHPUnit.

Even though it is not strictly part of this story, please install `symfony/phpunit-bridge` as a dev dependency. Once installed, we are automatically informed after each test run, if we still use stuff, which is considered deprecated.

As of PHPUnit 9.3 for example, [the `->at()` method has been deprecated](https://github.com/sebastianbergmann/phpunit/issues/4297), which we should take care of.

TheKeymaster (changes requested):

composer.json
RimantasAdeoWeb: Omg why would they deprecate it...
"it can lead to brittle tests which are too closely tied to specific implementation details" my ass, when the suggested alternative is specifying the entire sequence of params and returns anyway.

composer.json
```diff
@@ -21,16 +21,17 @@
...
     "require-dev": {
-        "phpunit/phpunit": "^9.0",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/var-dumper": "^5.1",
         "symfony/console": "^5.1",
         "symfony/filesystem": "^5.1",
-        "symfony/finder": "^5.1"
+        "symfony/finder": "^5.1",
+        "phpunit/phpunit": "^9.5",
+        "symfony/phpunit-bridge": "^5.2"
```
TheKeymaster, line 30: :raised_hands:

TheKeymaster (approved):

composer.json
TheKeymaster: Changing it like that makes you force writing cleaner code. It can be quite tricky for an existent code base with bad code, there will be hard times refactoring a simple `->at()` to a big `->exactly(xx)->withConsecutive()`.

But we are lucky, as we only used `->at()` twice in our code, and having a good code base :+1:!

PR #30, opened 2021-05-27
Plenty 339

RimantasAdeoWeb:


src/Wrapper/PropertyAware.php
```diff
@@ -90,10 +90,11 @@ protected function getPropertyValue(PimProperty $property)
 
                 return $this->registryService->getPropertySelections()->getPropertySelectionValues(
                     $property->getId(),
+                    $property->getValues(),
                     $this->config->getLanguage()
                 );
             default:
-                return $property->getPropertyData()->getNames()[0]->getValue() ?? null;
+                return $property->getValues()[0]->getValue();
```
TheKeymaster, line 97: I would still prefer having a fallback, in case it really does not exist.

```php
return isset($property->getValues()[0]) ? $property->getValues()[0]->getValue() : null;
```

tests/Wrapper/VariationTest.php
```diff
@@ -357,8 +357,8 @@ public function testNonExportablePropertiesAreSkipped(
     public function testPropertiesAreProperlyExported(): void
     {
         $expectedExportedAttributes = [
-            new Attribute('Color', ['Bereichsslider']),
-            new Attribute('Color', ['Hunde']),
+            new Attribute('Color', ['8']),
+            new Attribute('Color', ['100']),
```
TheKeymaster, line 361: Can these be changed to real values? Color `8` or `100` does not really make sense for unit-tests.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #31, opened 2021-06-07
PLENTY-319 ability to export variations with groupable attributes as separate items

RimantasAdeoWeb:


src/Wrapper/CsvWrapper.php
```diff
@@ -118,4 +113,78 @@ public function getExportPath(): string
...
     }
+
+    // TODO: come up with a better method name
+    /**
+     * @param PimVariation[] $productVariations
+     */
+    private function doWrap(
+        ProductEntity $product,
+        array $productVariations,
+        bool $separateVariationMode = false
```
TheKeymaster, line 124: I prefer not having such boolean parameters. Why not have a `int $mode` parameter? We could then simply say `Mode::DEFAULT` and `Mode::SEPARATE_VARIATIONS`. This way we could also add more modes in the future.

src/Wrapper/CsvWrapper.php
```diff
@@ -118,4 +113,78 @@ public function getExportPath(): string
...
+                $attribute = $this->registryService->getAttribute($attributeValue->getId());
+
+                if ($attribute && $attribute->isGroupable()) {
+                    $separateVariations[] = $variation;
+                    continue 2;
+                }
+            }
+
+            $groupedVariations[] = $variation;
+        }
```
TheKeymaster, line 175: Can we simplify this a bit? I need to understand both loops in order to know why we have `continue 2` here.

src/Wrapper/CsvWrapper.php
```diff
@@ -118,4 +113,78 @@ public function getExportPath(): string
...
+            return null;
+        }
+
+        return $item;
+    }
+
+    /**
+     * @param Variation[] $productVariations
+     */
+    private function splitVariationsByGroupability(array $productVariations): array
```
TheKeymaster, line 153: Maybe we can add a bit of context in the docblock, which describes the behavior a bit better.

src/Wrapper/CsvWrapper.php
```diff
@@ -118,4 +113,78 @@ public function getExportPath(): string
...
+
+    private function shouldExportGroupableAttributeVariantsSeparately(): bool
+    {
+        $config = $this->registryService->getPluginConfigurations('Ceres');
+
+        if (!isset($config['item.variation_show_type'])) {
+            return true;
+        }
+
+        return $config['item.variation_show_type'] == 'all';
```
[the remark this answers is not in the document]
TheKeymaster: Please also put `all` as a separate constant somewhere.

src/Wrapper/CsvWrapper.php
```diff
@@ -71,27 +74,19 @@ public function wrap(
...
-                    'Product with id %d could not be exported. Reason: %s',
-                    $product->getId(),
-                    $productWrapper->getReason()
-                ));
-
-                continue;
+            list($groupedVariations, $separateVariations) = $this->splitVariationsByGroupability($productVariations);
+
+            if ($groupedVariations) {
+                if ($item = $this->doWrap($product, $groupedVariations)) {
```
TheKeymaster, line 80: Can we have a bit more defensive coding here? Nested `if`s are kind of hard to get into.

src/Wrapper/Product.php
```diff
@@ -46,6 +46,11 @@ class Product
     /** @var Exporter */
     private $exporter;
 
+    /**
+     * @var bool
+     */
+    private $separateMode;
```
TheKeymaster, line 52: We use PHP 7.4, so this could be properly typehinted as `int` (see my comment regarding modes).

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -0,0 +1,554 @@
...
+            null,
+            $this->exporterMock,
+            $this->config,
+            $this->registryServiceMock,
+            $this->loggerMock,
+            $this->loggerMock
+        );
+    }
+
+    public function testItExportsEachVariationSeparatelyIfConfiguredToDoSoAndIfAllVariationsHaveAGroupableAttribute()
```
TheKeymaster, line 102: We can get rid of `It` as a test always describes "`a`" behavior. Also `ToDoSo` isn't really a clear description of what the test should actually do. Why not call it something like `testGroupableVariantsAreExportedAsSeparateItemsWhenConfigured`. Maybe you can come up with a name that is shorter or even more expressive.

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -0,0 +1,554 @@
...
+
+        $attributeResponse = $this->getMockResponse('AttributeResponse/response.json');
+        $attributes = AttributeParser::parse($attributeResponse);
+        $this->registryServiceMock->method('getAttribute')->with(3)->willReturn($attributes->findOne(['id' => 3]));
+
+        $this->exporterMock->expects($this->once())
+            ->method('serializeItemsToFile')
+            ->with(
+                self::TEST_EXPORT_PATH,
+                $this->callback(function ($items) {
```
TheKeymaster, line 135: You can actually typehint `array $items`

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -0,0 +1,554 @@
...
+                    }
+
+                    return true;
+                })
+            );
+
+        $this->csvWrapper->wrap(0, 1, $items, $variations);
+    }
+
+    public function testItExportsTwoItemsWhenOnlyOneOutOfThreeVariantsHasGroupableAttributes()
```
TheKeymaster, line 242: Please see my comment above regarding test names. This applies to all tests in this class.

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -0,0 +1,554 @@
...
+    }
+
+    public function testItGroupsVariantsWithGroupableAttributesIntoASingleItemIfConfiguredNotToShowSeparately()
+    {
+        $this->registryServiceMock->method('getPluginConfigurations')
+            ->with('Ceres')
+            ->willReturn(
+                [
+                    'global.enableOldUrlPattern' => false,
+                    'item.variation_show_type' => 'notAll'
```
TheKeymaster, line 308: Is it really `notAll`? In my other comment I asked to add a constant for `all` so it probably also makes sense to create one for `notAll` as well (if this is actually how Plenty named this).

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -0,0 +1,554 @@
...
+                $attributes->findOne(['id' => 3]),
+                $attributes->findOne(['id' => 3]),
+                $attributes->findOne(['id' => 3])
+            );
+
+        $this->exporterMock->expects($this->once())
+            ->method('serializeItemsToFile')
+            ->with(
+                self::TEST_EXPORT_PATH,
+                $this->callback(function ($items) {
```
TheKeymaster, line 336: See my other comment regarding the callback function type hints. This applies to all newly added callbacks.

TheKeymaster (changes requested):

tests/Wrapper/CsvWrapperTest.php
RimantasAdeoWeb: Naw, they have multiple options there, there's no actual 'notAll'. Do I make this 'anyConfigValueAsLongAsItIsNotAll' to make it clearer?

TheKeymaster: Can't we just use the option provided by Plenty if it is indeed not `all`, what would this be and can we use that?

PR #32, opened 2021-06-22
PLENTY-344 handle tag for export exclusion

RimantasAdeoWeb:


src/Response/Entity/Pim/Variation.php
```diff
@@ -19,6 +19,8 @@
 
 class Variation extends Entity
 {
+    public const EXCLUSION_TAG = 'findologic-exclude';
```
TheKeymaster, line 22: We may want to change that naming a bit.

```suggestion
    public const EXCLUSION_TAG_NAME = 'findologic-exclude';
```

src/Response/Entity/Pim/Variation.php
```diff
@@ -193,4 +195,21 @@ public function getUnit(): ?Unit
     {
         return $this->unit;
     }
+
+    public function hasExportExclusionTag(string $lang): bool
+    {
+        foreach ($this->getTags() as $tag) {
+            foreach ($tag->getTagData()->getNames() as $name) {
```
TheKeymaster, line 202: For the future we may want to think about having a `Nameable`/`Translatable` interface, which would allow us to simply get the translated name of a given entity with some helper method. Not an issue now though.

src/Response/Entity/Pim/Variation.php
```diff
@@ -193,4 +195,21 @@ public function getUnit(): ?Unit
     {
         return $this->unit;
     }
+
+    public function hasExportExclusionTag(string $lang): bool
```
TheKeymaster, line 199: Neat helper method. Also makes sense to have this available directly at the variation :+1:.

src/Wrapper/CsvWrapper.php
```diff
@@ -70,6 +70,17 @@ public function wrap(
         /** @var Item[] $items */
         $items = [];
         foreach ($products->all() as $product) {
+            if (!$this->shouldExportProduct($product, $variations)) {
+                $this->customerLogger->alert(
```
TheKeymaster, line 74: I think an `alert` is a bit over kill here. I would suggest a `notice` or `info` as the customer probably has assigned this tag for a reason and is aware of it.

src/Wrapper/CsvWrapper.php
```diff
@@ -199,4 +210,20 @@ private function shouldExportGroupableAttributeVariantsSeparately(): bool
...
+    private function shouldExportProduct(ProductResponseItem $product, PimVariationResponse $variations): bool
+    {
+        $mainVariation = $variations->findOne([
+            'base' => [
+                'itemId' => $product->getId(),
+                'isMain' => true
+            ]
+        ]);
+
+        if ($mainVariation && $mainVariation->hasExportExclusionTag($this->config->getLanguage())) {
```
TheKeymaster, line 223: This is so nice to read

image

TheKeymaster (changes requested):

src/Wrapper/CsvWrapper.php
RimantasAdeoWeb: Not entirely sure what's special about this, but glad to hear that nonetheless :D

TheKeymaster (approved):

PR #33, opened 2021-07-21
Plenty 347

RimantasAdeoWeb:


src/Response/Collection/PropertySelectionResponse.php
```diff
@@ -92,14 +94,15 @@ public function getPropertySelectionValues(int $propertyId, array $selections, s
...
 
         $values = [];
         foreach ($propertySelections as $propertySelection) {
-            foreach ($propertySelection->getRelation()->getRelationValues() as $relationValue) {
-                if (strtoupper($relationValue->getLang()) == strtoupper($lang)) {
-                    $values[$propertySelection->getId()] = $relationValue->getValue();
-                }
+            /** @var RelationValue|null $relationValue */
+            $relationValue = Translator::translate($propertySelection->getRelation()->getRelationValues(), $lang);
+            if ($relationValue) {
```
RimantasAdeoWeb, line 104: Couldn't use `if ($relationValue = Translator:...)` syntax because apparently PHPStan can't connect docblocks to such variable declarations.

src/Parser/PropertySelectionParser.php
```diff
@@ -25,12 +25,12 @@ public static function parse(ResponseInterface $rawResponse): Response
 
         return new PropertySelectionResponse(
             $response['page'],
-            $response['totalsCount'],
+            (int)$response['totalsCount'],
```
TheKeymaster, line 28: Do we need those casts for other response instances as well?

src/Response/Collection/PropertySelectionResponse.php
TheKeymaster: :+1:

src/Wrapper/CsvWrapper.php
```diff
@@ -78,7 +78,7 @@ public function wrap(
                     )
                 );
 
-                return;
+                continue;
```
TheKeymaster, line 81: Nice catch! Can we add a unit-test, so we don't remove this again by accident in the future?

src/Wrapper/Variation.php
```diff
@@ -376,12 +379,10 @@ private function processTags(): void
...
-
-                    break;
-                }
+            /** @var TagName|null $translatedTag */
+            $translatedTag = Translator::translate($tag->getTagData()->getNames(), $this->config->getLanguage());
+            if ($translatedTag) {
+                $tagName = $translatedTag->getName();
             }
 
             $this->tags[] = new Keyword($tagName);
```
TheKeymaster, line 388: Wouldn't this cause an exception when `$translatedTag` returns `null`, as `$tagName` is never declared?

TheKeymaster (changes requested):

src/Wrapper/Variation.php
RimantasAdeoWeb: $tagName is declared a few lines above.

src/Parser/PropertySelectionParser.php
RimantasAdeoWeb: No, probably a bug in the Plentymarkets API with this endpoint.

TheKeymaster (approved):

PR #35, opened 2021-08-05
PLENTY-349 exportable ordernumber fields configuration

RimantasAdeoWeb:


src/Config.php
```diff
@@ -11,38 +11,37 @@
  */
 class Config
 {
-    /** @var string */
-    private $domain;
+    private string $domain;
```
TheKeymaster, line 14: Great idea utilizing the typed properties, since we upgraded to 7.4.

src/Config.php
```diff
@@ -231,4 +240,54 @@ public function setExportUnavailableVariations(bool $exportUnavailableVariations
     {
         $this->exportUnavailableVariations = $exportUnavailableVariations;
     }
+
+    public function getExportOrdernumberProductId(): bool
+    {
+        return $this->exportOrdernumberProductId;
+    }
+
+    public function setExportOrdernumberProductId(bool $flag): void
```
TheKeymaster, line 249: I prefer having the name of the property as the name of the argument rather than calling it `$flag`.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #37, opened 2021-08-17
PLENTY-360_fix_property_bugs_and_add_error_handling

TheKeymaster:


tests/ClientTest.php
```diff
@@ -244,16 +244,16 @@ public function testRateLimitIsHandled(): void
...
             );
 
         $request = new WebStoreRequest();
         $client->send($request);
-        $expectedRateLimitedTime = Carbon::now()->addSeconds(5);
+        $expectedRateLimitedTime = Carbon::now()->addSeconds(2);
         $client->send($request);
 
-        $this->assertEqualsWithDelta($expectedRateLimitedTime->timestamp, Carbon::now()->timestamp, 2);
+        $this->assertEqualsWithDelta($expectedRateLimitedTime->timestamp, Carbon::now()->timestamp, 1);
```
TheKeymaster, line 256: 2 seconds of waiting for this test is sufficient. 5 seconds is way to long and slowed down the whole test-suite, which is quite annoying when running the whole suite locally multiple times.

tests/Exporter/ExporterTest.php
```diff
@@ -83,6 +85,13 @@ protected function setUp(): void
             ->willReturn($parsedCategoryResponse->first());
     }
 
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        $_ENV['APP_ENV'] = 'test';
```
dblaichinger, line 92: Why do we need to set it in tearDown? Shouldn't this be done in setUp?

dblaichinger (changes requested):

tests/Exporter/ExporterTest.php
TheKeymaster: Here this is also from the PHPUnit documentation:

image

dblaichinger (approved):

PR #39, opened 2021-09-01
PLENTY-361 group wrap failure logs by failure reason

RimantasAdeoWeb:


src/Wrapper/CsvWrapper.php
```diff
@@ -105,6 +101,8 @@ public function wrap(
         }
 
         $this->exporter->serializeItemsToFile($this->exportPath, $items, $start, count($items), $total);
+
+        $this->logFailures();
```
TheKeymaster, line 105: Please also group skipped products (line 70-75).

TheKeymaster: Also add a separate test to ensure they're also grouped.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #40, opened 2021-09-14
Plenty 369

RimantasAdeoWeb:


src/Response/Collection/PropertyGroupResponse.php
```diff
@@ -19,14 +19,14 @@ class PropertyGroupResponse extends IterableResponse implements CollectionInterf
...
+        bool  $isLastPage,
         array $propertyGroups,
-        int $lastPageNumber = 1,
-        int $firstOnPage = 1,
-        int $lastOnPage = 1,
-        int $itemsPerPage = 100
+        int   $lastPageNumber = 1,
+        int   $firstOnPage = 1,
+        int   $lastOnPage = 1,
+        int   $itemsPerPage = 100
```
TheKeymaster, line 29: These indents were probably added by accident?

src/Response/Entity/ItemPropertyGroup/Name.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /** @var string|null */
+    private $name;
+
+    /** @var string|null */
+    private $description;
+
+    public function __construct(array $data)
+    {
+        $this->propertyGroupId = (int)$data['propertyGroupId'];
```
TheKeymaster, line 26: Can't we use `getIntProperty` here?

src/Response/Entity/ItemPropertyGroup/Name.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function getPropertyGroupId(): int
+    {
+        return $this->propertyGroupId;
+    }
+
+    public function getLang(): string
+    {
+        return (string)$this->lang;
```
TheKeymaster, line 49: Wouldn't this be more explicit?

```suggestion
        return $this->lang ?? '';
```

src/Wrapper/PropertyAware.php
```diff
@@ -45,6 +45,21 @@ public function processProperties(): void
             }
 
             $value = $this->getPropertyValue($property);
+            if (!$value && $property->getPropertyData()->getCast() == CastType::EMPTY) {
```
TheKeymaster, line 48: I know in the old exporter we weren't strict about this, but we should change that, because it is a string comparison anyway

```suggestion
            if (!$value && $property->getPropertyData()->getCast() === CastType::EMPTY) {
```

src/Wrapper/PropertyAware.php
```diff
@@ -67,35 +82,28 @@ protected function getPropertyValue(PimProperty $property)
...
-                        continue;
-                    }
-
-                    $relation = $selection->getRelation();
-                    if (!$relation) {
-                        continue;
-                    }
+                $value = $this->registryService->getPropertySelections()->getPropertySelectionValues(
+                    $property->getId(),
+                    [$property->getValues()[0]],
```
TheKeymaster, line 101: Are we sure that the first index will always exist here?

src/Wrapper/PropertyAware.php
```diff
@@ -67,35 +82,28 @@ protected function getPropertyValue(PimProperty $property)
...
-                foreach ($property->getPropertyData()->getSelections() as $selection) {
-                    if ($property->getValues()[0]->getValue() != $selection->getId()) {
-                        continue;
-                    }
-
-                    $relation = $selection->getRelation();
-                    if (!$relation) {
-                        continue;
-                    }
+                $value = $this->registryService->getPropertySelections()->getPropertySelectionValues(
```
TheKeymaster, line 99: We can be explicit about the naming, because `$value` isn't very descriptive. I need to check the method name in line 77 to know for what this value is used, but I would still have no idea where this is coming from, when I am checking this from line 106.

```suggestion
                $propertySelectionValues = $this->registryService->getPropertySelections()->getPropertySelectionValues(
```

This makes it also clearer, why we are using `reset($value)` at the end, as it contains an array of selection values.

tests/Response/Collection/ItemPropertyGroupResponseTest.php
```diff
@@ -0,0 +1,92 @@
...
+    use ResponseHelper;
+
+    private $response;
+
+    private $propertyGroupResponse;
+
+    public function setUp(): void
+    {
+        // Set timezone for proper comparison of timestaps in the export.
+        date_default_timezone_set('Europe/Vienna');
```
TheKeymaster, line 23: Please set this in the boostrap of all tests (`tests/bootstrap.php`), as this should be either done for all or for no tests. If you don't want this, we should then ensure to reset this to how it was before in `tearDown`, as otherwise this may impact some tests and not others, which you will only notice much later down the line, when you already spent endless time debugging.

tests/Response/Collection/ItemPropertyGroupResponseTest.php
```diff
@@ -0,0 +1,92 @@
...
+     * @dataProvider criteriaProvider
+     */
+    public function testCriteriaSearchWorksAsExpected(array $criteria, int $expectedId): void
+    {
+        $propertyGroup = $this->propertyGroupResponse->findOne($criteria);
+
+        $this->assertEquals($expectedId, $propertyGroup->getId());
+    }
+
+    public function testGetAllReturnsCorrectNumberOfItems()
```
TheKeymaster, line 50: Please check the tests to always include a typehint.

tests/Response/Collection/PropertyGroupResponseTest.php
```diff
@@ -19,8 +19,6 @@ class PropertyGroupResponseTest extends TestCase
 
     public function setUp(): void
     {
-        // Set timezone for proper comparison of timestaps in the export.
-        date_default_timezone_set('Europe/Vienna');
```
TheKeymaster, line 23: I guess this was the issue that I was talking about in my previous comment. It's better to move this in the bootstrap.

TheKeymaster (changes requested):

src/Wrapper/PropertyAware.php
RimantasAdeoWeb: Yes

TheKeymaster (approved):

PR #49, opened 2021-11-22
PLENTY-393 skipping category if parent category is not visible

msilinskas:


src/Wrapper/Variation.php
```diff
@@ -279,9 +285,13 @@ private function buildCategoryPath(Category $category): string
 
         if ($categoryDetail = $this->getCategoryDetailForCurrentPlentyIdAndLanguage($category)) {
             if ($category->getParentCategoryId() !== null) {
-                $path[] = $this->buildCategoryPath(
-                    $this->registryService->getCategory($category->getParentCategoryId())
-                );
+                $parentCategory = $this->registryService->getCategory($category->getParentCategoryId());
+
+                if (Utils::isEmpty($parentCategory)) {
```
TheKeymaster, line 290: `Utils::isEmpty` is really helpful for values that we get from the API and we want to add them to our products data. Here however the return value is either an instance of the `Category` class, or `null`, so simply checking for not nullable or even a falseness check should be sufficient.

src/Wrapper/Variation.php
```diff
@@ -279,9 +285,13 @@ private function buildCategoryPath(Category $category): string
...
+                    return '';
```
TheKeymaster, line 291: I would prefer to make this method nullable. This way it is clear what happens in case the category path can not be built for whatever reason, and it is also easy to check e.g. in line 269.

tests/Wrapper/VariationTest.php
```diff
@@ -205,6 +205,32 @@ public function testChildCategoriesAreProperlyBuilt(): void
         $this->assertSame('/living-room/armchairs-stools/', $wrapper->getAttributes()[1]->getValues()[0]);
     }
 
+    public function testCategoryIsSkippedIfParentCategoryIsNotVisible(): void
```
TheKeymaster, line 208: Add a bit more context, so it is clear how a category can be "not visible".

```suggestion
    public function testCategoryIsSkippedIfParentCategoryIsNotVisibleForExportingClient(): void
```

TheKeymaster (changes requested):

src/Wrapper/Variation.php
TheKeymaster: This was resolved without the fix applied. Did you forget to commit the change?

TheKeymaster (changes requested):

src/Wrapper/Variation.php
msilinskas: I've forgot to commit this change, sorry

TheKeymaster (approved):

PR #53, opened 2021-12-03
PLENTY-400 fixing image uri

msilinskas:


src/Wrapper/CsvWrapper.php
```diff
@@ -84,8 +84,15 @@ public function wrap(
...
             list($groupedVariations, $separateVariations) = $this->splitVariationsByGroupability($productVariations);
 
-            foreach ($separateVariations as $separateVariation) {
-                if ($item = $this->wrapItem($product, $separateVariation, Product::WRAP_MODE_SEPARATE_VARIATIONS)) {
+            foreach ($separateVariations as $key => $separateVariation) {
+                $item = $this->wrapItem(
+                    $product,
+                    $separateVariation,
+                    Product::WRAP_MODE_SEPARATE_VARIATIONS,
+                    (string)$key
```
TheKeymaster, line 92: Why is this casted to a string?

src/Wrapper/Product.php
```diff
@@ -65,7 +67,8 @@ public function __construct(
         RegistryService $registryService,
         ProductEntity $productEntity,
         array $variationEntities,
-        int $wrapMode = self::WRAP_MODE_DEFAULT
+        int $wrapMode = self::WRAP_MODE_DEFAULT,
+        string $variationAttributes = ''
```
TheKeymaster, line 71: The name of this variable suggests an array, but it really is a string. Please try to find a better fitting name and rename it in other places accordingly.

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+                    $imageAttributeValues = $image->getAttributeValueImages();
+
+                    if ($this->checkImageAvailability($imageAttributeValues, $variationAttributes)) {
+                        $this->image = new Image($image->getUrlMiddle());
+
+                        return;
+                    }
+                }
+            }
+        }
```
TheKeymaster, line 465: This is nested in a way which makes it honestly really hard to read. Try to exit early and apply some defensive coding, instead of nesting if blocks in further if blocks.

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+    }
+
+    private function checkImageAvailability(array $imageAttributeValues, array $variationAttributes): bool
```
TheKeymaster, line 468: A method that returns a boolean typically indicates a state of something. Such as `isAvailable` or `hasValues`. Without checking the method I wouldn't really know what it "checks", so please try to find a better fitting name.

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+
+            if ($this->image !== null) {
+                return;
+            }
+        }
+
+        $this->setImageAvailabilityForDefaultWrapMode($images);
+    }
+
+    private function setImageForSeparatedVariation(array $images, array $variationAttributes): void
```
TheKeymaster, line 450: When using the `array` type always add the structure of it to the docblock. Please do this for all methods.

```php
/**
 * @param Entity[] $entities
 */
public function someMethod(array $entities): void
{
    // ...
}
```

The only exception to this rule is where an array structure may be too complex, where a docblock serves not much purpose, such as:

image

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+                $imageAttributeId = $imageAttributeValue->getAttributeId();
+                $imageValueId = $imageAttributeValue->getValueId();
+
+                if ($imageAttributeId == $attributeData[0] && $imageValueId == $attributeData[1]) {
+                    $count++;
+                }
+            }
+        }
+
+        return $count === count($variationAttributes);
```
TheKeymaster, line 484: Without knowing deep internals, I would never know what this does. Try to move all things related to the separate variation logic into a separate class, which is responsible for it. Right now the code seems to creep up everywhere.

tests/Response/Collection/PimVariationResponseTest.php
```diff
@@ -189,12 +188,15 @@ public function testPimVariationDataCanBeFetched(): void
         unset($expectedImageData['availabilities']);
         unset($expectedImageData['createdAt']);
         unset($expectedImageData['updatedAt']);
+        unset($expectedImageData['updatedAt']);
```
TheKeymaster, line 191: Unsetting this twice won't do much :D

tests/Wrapper/CsvWrapperTest.php
```diff
@@ -822,11 +822,19 @@ public function testSeparatedVariationsIsGroupedBasedOnTwoGroupableAttributes():
...
+                        'https://cdn03.plentymarkets.com/item/images/108/middle/108-Barsessel-Black-Mamba-3.jpg',
+                        'https://cdn03.plentymarkets.com/item/images/108/middle/108-Barsessel-Black-Mamba-4.jpg'
+                    ];
+
                     foreach ($items as $key => $item) {
                         $line = $item->getCsvFragment();
                         $columnValues = explode("\t", $line);
                         $this->assertEquals($expectedIds[$key], $columnValues[0]);
                         $this->assertEquals($expectedOrderNumbers[$key], $columnValues[1]);
+                        $this->assertEquals($expectedImagesUrls[$key], $columnValues[10]);
```
TheKeymaster, line 837: When the type is clear, I typically prefer `assertSame`, but this is also fine.

src/Wrapper/Variation.php
TheKeymaster: Maybe this foreach logic can somehow be centeralized, as it is very similar to `setImageAvailabilityForDefaultWrapMode`.

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+    {
+        $count = 0;
+
+        foreach ($imageAttributeValues as $imageAttributeValue) {
+            foreach ($variationAttributes as $variationAttribute) {
+                $attributeData = explode('_', $variationAttribute);
+                $imageAttributeId = $imageAttributeValue->getAttributeId();
+                $imageValueId = $imageAttributeValue->getValueId();
+
+                if ($imageAttributeId == $attributeData[0] && $imageValueId == $attributeData[1]) {
```
TheKeymaster, line 478: Why aren't you using strict comparison (`===`)?

src/Wrapper/Variation.php
```diff
@@ -420,11 +431,61 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+                    $count++;
+                }
+            }
+        }
+
+        return $count === count($variationAttributes);
+    }
+
+    private function setImageAvailabilityForDefaultWrapMode(array $images): void
```
TheKeymaster, line 487: This method name would make me assume it is setting some availability for images, but instead it is actually setting the first image which is available for the store.

src/Wrapper/Variation.php
TheKeymaster: This class should also contain the logic to build the key, parsing the key and generally everything related to those separated variations.

TheKeymaster (changes requested):

src/Wrapper/CsvWrapper.php
msilinskas: Sometimes key is a product id, which has a type of int

src/Wrapper/Variation.php
msilinskas: Sometimes exploded data has different type than data fetched from the image object.

src/Wrapper/Variation.php
msilinskas: This method was removed and the logic was nested with the separated vartiations logic.

src/Wrapper/Product.php
```diff
@@ -68,7 +68,7 @@ public function __construct(
         ProductEntity $productEntity,
         array $variationEntities,
         int $wrapMode = self::WRAP_MODE_DEFAULT,
-        string $variationAttributes = ''
+        string $variationAttributeKey = ''
```
TheKeymaster, line 71: I would rather name this `$variationGroupKey`, as it references a grouped variation.

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                $imageAvailable = true;
+            }
+        }
+
+        return $imageAvailable;
+    }
+
+    public function getVariationGroupKey(): string
```
TheKeymaster, line 53: Please add a docblock of how these may look like for different scenarios and also elaborate on how this structure is being made.

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+        Variation $variation,
+        RegistryService $registryService
+    ) {
+        $this->variation = $variation;
+        $this->registryService = $registryService;
+    }
+
+    /**
+     * @param ImageAttributeValue[] $imageAttributeValues
+     * @param array $variationAttributes
```
TheKeymaster, line 27: Isn't his a key/value of strings?

```suggestion
     * @param array<string, string> $variationAttributes
```

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace FINDOLOGIC\PlentyMarketsRestExporter\Wrapper;
+
+use FINDOLOGIC\PlentyMarketsRestExporter\RegistryService;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Entity\Pim\Property\ImageAttributeValue;
+use FINDOLOGIC\PlentyMarketsRestExporter\Response\Entity\Pim\Variation;
+
+class SeparatedVariation
+{
+    private Variation $variation;
```
TheKeymaster, line 13: I typically prefer no spaces between class properties when they're strictly typed by PHP. But that's just preference.

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+
+        foreach ($imageAttributeValues as $imageAttributeValue) {
+            foreach ($variationAttributes as $variationAttribute) {
+                $attributeData = explode('_', $variationAttribute);
+                $imageAttributeId = $imageAttributeValue->getAttributeId();
+                $imageValueId = $imageAttributeValue->getValueId();
+                $variationAttributeId = $attributeData[0];
+                $variationValueId = $attributeData[1];
+
+                if ($imageAttributeId != $variationAttributeId || $imageValueId != $variationValueId) {
```
TheKeymaster, line 42: Why aren't you using a strict comparison?

```suggestion
                if ($imageAttributeId !== $variationAttributeId || $imageValueId !== $variationValueId) {
```

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+
+                if ($key !== '') {
+                    $key .= '/';
+                }
+
+                $key .= $attributeValue->getId() . '_' . $attributeValue->getValue()->getId();
+            }
+        }
+
+        return ($groupableAttributes === 0 || count($attributeValues) > 1) ? $key : (string)$this->variation->getId();
```
TheKeymaster, line 73: This might be more readable with an if.

```suggestion
        if ($groupableAttributes === 0 || count($attributeValues) > 1)   {
            return $key;
        }

        return (string)$this->variation->getId();
```

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+    public function getVariationGroupKey(): string
+    {
+        $attributeValues = $this->variation->getAttributeValues();
+        $groupableAttributes = 0;
+        $key = '';
+
+        foreach ($attributeValues as $attributeValue) {
+            $attribute = $this->registryService->getAttribute($attributeValue->getId());
+
+            if ($attribute && $attribute->isGroupable()) {
```
TheKeymaster, line 62: This may be easier readable by skipping early. This also prevents those nested if blocks.
```suggestion
            if (!$attribute || !$attribute->isGroupable()) {
                continue;
            }
```

src/Wrapper/Variation.php
```diff
@@ -420,21 +431,43 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
 
                     return;
                 }
+
+                if (!$defaultWrapModeImage) {
+                    $defaultWrapModeImage = $image;
+                }
+
+                $separatedVariation = new SeparatedVariation($this->variationEntity, $this->registryService);
+
```
TheKeymaster, line 457: Maybe this line can be removed to move the structurally dependent things more together.

src/Wrapper/Variation.php
```diff
@@ -420,21 +431,43 @@ private function shouldProcessTag(Tag $tag, int $storeId): bool
...
+                if (!$separatedVariation->isImageAvailable($image->getAttributeValueImages(), $variationAttributes)) {
+                    continue;
+                }
+
+                $this->image = new Image($image->getUrlMiddle());
+
+                return;
```
TheKeymaster, line 464: Those returns after setting the image instances seem a bit wrong. Maybe you want to add a method, which is responsible for getting the image. `processImages` would then be simply responsible for setting it to this class property.

```suggestion
    private function processImages(): void
    {
        $this->image = $this->getVariantImage();
    }
```

src/Wrapper/Variation.php
```diff
@@ -96,16 +97,26 @@ class Variation
     /** @var string|null */
     protected $packageSize;
 
+    /** @var string */
+    protected $variationAttributesKey;
```
TheKeymaster, line 101: These class properties can also be typed.

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+                $groupableAttributes++;
+
+                if ($key !== '') {
+                    $key .= '/';
```
TheKeymaster, line 66: This feels like it wants to be a constant.

src/Wrapper/SeparatedVariation.php
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                $key .= $attributeValue->getId() . '_' . $attributeValue->getValue()->getId();
```
TheKeymaster, line 69: `_` feels like it wants to be a constant.

TheKeymaster (changes requested):

src/Wrapper/Variation.php
TheKeymaster: As already noted, maybe a docblock would make it easier to understand.

src/Wrapper/Variation.php
TheKeymaster: Wouldn't it make more sense to add a typecheck or cast the values to their respective type?

src/Wrapper/SeparatedVariation.php
msilinskas: Nop, it hasa value like [0] => '112_150', [1] => '113_140'

src/Wrapper/Variation.php
msilinskas: We need to stop the execution when image is set to an object

src/Wrapper/Variation.php
msilinskas: I've added a docblock.

src/Wrapper/Variation.php
msilinskas: I've added a type casts for the values

src/Wrapper/SeparatedVariation.php
TheKeymaster: My point is, if we know what values it may contain, we should definitely typehint it. So it would be `array<int, string>` ot `string[]` then.

src/Wrapper/Variation.php
TheKeymaster: Yes this is why I would like this to be a method that returns it instead of setting it and returning and therefore stopping further execution. Maybe this example makes it a bit more clear :D

```php
private function processImages(): void
{
    $this->image = $this->getVariantImage();
}

private function getVariantImage(): Image
{
    // foreach ...
    if ($this->wrapMode === Product::WRAP_MODE_DEFAULT) {
        return new Image($image->getUrlMiddle());
    }
    // ...
}
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #55, opened 2021-12-20
PLENTY-407 translating characteristics

msilinskas:


src/RegistryService.php
```diff
@@ -392,7 +392,8 @@ protected function fetchPropertyGroups(): void
 
     protected function fetchItemProperties(): void
     {
-        $propertyRequest = new ItemPropertyRequest();
+        $with = 'names,selections';
+        $propertyRequest = new ItemPropertyRequest($with);
```
TheKeymaster, line 396: Can we make it so this expects an array? The `Client` should be able to automatically map those by concatenating all arrays with the `,` character.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #57, opened 2022-01-11
PLENTY-416_variation_id_in_url_depending_on_config

dblaichinger:


src/Wrapper/Product.php
```diff
@@ -79,6 +81,7 @@ public function __construct(
         $this->storeConfiguration = $storeConfiguration;
         $this->wrapMode = $wrapMode;
         $this->variationGroupKey = $variationGroupKey;
+        $this->plentyShopConfig = $this->registryService->getPluginConfigurations('Ceres');
```
TheKeymaster, line 84: Can we change the existing calls in this class to now use the private property instead?

src/Wrapper/Product.php
```diff
@@ -366,36 +369,54 @@ protected function addFreeTextFields(): void
...
+            $this->getWebStoreHost(),
+            $this->getLanguageUrlPrefix(),
+            trim($urlPath, '/'),
+            $this->productEntity->getId(),
+        );
 
-        if (!isset($config['global.enableOldUrlPattern'])) {
-            return true;
+        if (isset($this->plentyShopConfig['item.show_please_select'])) {
+            if (filter_var($this->plentyShopConfig['item.show_please_select'], FILTER_VALIDATE_BOOLEAN)) {
```
TheKeymaster, line 411: I think there is already another place where we use this `filter_var`. Can we generalize this in a static helper method, maybe in `Utils`?

tests/Wrapper/ProductTest.php
```diff
@@ -847,6 +847,85 @@ public function testCallistoUrlFormatIsUsedWhenConfigured(): void
         );
     }
 
+    public function itemShowPleaseSelectProvider()
```
TheKeymaster, line 850: Add missing typehint

```suggestion
    public function itemShowPleaseSelectProvider(): array
```

tests/Wrapper/ProductTest.php
```diff
@@ -847,6 +847,85 @@ public function testCallistoUrlFormatIsUsedWhenConfigured(): void
...
+            ]
+        ];
+    }
+    /**
+     * @dataProvider itemShowPleaseSelectProvider
+     */
+    public function testPlentyShopUrlWithoutVariationIdIsUsedWhenConfigured(
+        $plentyShopConfig,
+        $baseUrlPath,
+        $expectedProductUrl
```
TheKeymaster, line 878: Add missing typehints.

tests/Wrapper/ProductTest.php
```diff
@@ -847,6 +847,85 @@ public function testCallistoUrlFormatIsUsedWhenConfigured(): void
         );
     }
 
+    public function itemShowPleaseSelectProvider()
+    {
+        $baseUrlPath = 'awesome-url-path/somewhere-in-the-store';
+        return [
+            'Url with variation id when "item.show_please_select" disabled' => [
```
TheKeymaster, line 854: I usually prefer having all data provider names being lower cased, as then, if the test would fail, everything is lower cased. But that's just my opinion.

image

tests/Wrapper/ProductTest.php
```diff
@@ -847,6 +847,85 @@ public function testCallistoUrlFormatIsUsedWhenConfigured(): void
...
+                'plentyShopConfig' => [
+                    'item.show_please_select' => true,
+                    'global.enableOldUrlPattern' => false,
+                ],
+                'baseUrlPath' => $baseUrlPath,
+                'expectedProductUrl' => $baseUrlPath . '_0'
+            ]
+        ];
+    }
+    /**
```
TheKeymaster, line 872: Please add a blank line between a new block/method.

TheKeymaster (changes requested):

src/Wrapper/Product.php
dblaichinger: `$this->plentyShopConfig` is used in the class, see `shouldUseCallistoUrl()` or `getPlentyShopUrl()`. Or what do you mean exactly?

TheKeymaster: Ah nevermind, the second call for this is being done in `CsvWrapper` and everything else already accesses this information via `$this->plentyShopConfig`.

TheKeymaster (approved):

PR #58, opened 2022-01-12
PLENTY-404 using cheapest variation information

msilinskas:


src/Wrapper/SeparatedVariation.php
```diff
@@ -57,16 +63,14 @@ public function isImageAvailable(array $imageAttributeValues, array $variationAt
      * and generates keys from groupable attributes ids and values. If there is just a one attribute,
      * then we returning a variation id for separating variations by one attribute.
      */
-    public function getVariationGroupKey(): string
+    public function getVariationGroupKey(bool $force = false, bool $checkIfGroupable = true): string
```
TheKeymaster, line 66: `$force` feels like the wrong name or generally maybe even the complete wrong approach. There are so many questions for this method argument;

* What are you forcing?
* Why are you forcing?
* When should you force?

I'd highly prefer a separate method for this, which should indicate clearly what this force parameter does, so you don't have to know the deep internals to have an idea what this `$force` method argument even stands for.

TheKeymaster, line 66: `$checkIfGroupable` is similar to `$force` a boolean. A good rule of thumb is to never, or only very rarely use booleans for method arguments, as when reading the method call itself it looks like `someMethod(true, false)`, which means you would need to go into the method to even check what causes this flag even. 

I would also suggest a separate method for this, so it is simpler and you can already read it directly. Like calling it `getVariationKeyForGroupedAttributes` or something similar. You also don't need to be aware of all these flags, that suddenly make the method do something, which isnt even covered by the method name itself, which just makes the whole method body unnecessarily complex.

src/Wrapper/SeparatedVariation.php
```diff
@@ -79,10 +83,30 @@ public function getVariationGroupKey(): string
...
+    public function getVariationAttributes(string $variationAttr, bool $checkIfGroupable): array
+    {
+        if (strpos($variationAttr, self::ATTRIBUTE_VALUE_SEPARATOR)) {
+            return explode(self::MULTIPLE_KEYS_SEPARATOR, $variationAttr);
+        }
+
+        return explode(self::MULTIPLE_KEYS_SEPARATOR, $this->getVariationGroupKey(true, $checkIfGroupable));
+    }
+
+    private function checkIfGroupable(Attribute $attributeValue): bool
```
TheKeymaster, line 102: Could simply be renamed to `isGroupable` or `isAttributeGroupable`. A "check" usually doesn't return anything, but may even throw an exception or return an `int`, which may refer to an exit code.

TheKeymaster: The naming of the method argument is kinda confusing, as the typehint suggests a whole `Attribute` instance, but the method argument is called `$attributeValue`.

src/Wrapper/SeparatedVariation.php
```diff
@@ -79,10 +83,30 @@ public function getVariationGroupKey(): string
...
+    {
+        $attribute = $this->registryService->getAttribute($attributeValue->getId());
+
+        if (!$attribute || !$attribute->isGroupable()) {
+            return false;
+        }
+
+        return true;
```
TheKeymaster, line 110: May be simplified to just returning it as is. I typically prefer to put braces around a boolean expression, but that's up to your code style.

```php
return (expression);
```

src/Wrapper/SeparatedVariation.php
```diff
@@ -79,10 +83,30 @@ public function getVariationGroupKey(): string
...
 
-        if ($groupableAttributes === 0 || count($attributeValues) > 1) {
+        if ($groupableAttributes === 0 || count($attributeValues) > 1 || $force) {
             return $key;
         }
 
         return (string)$this->variation->getId();
     }
+
+    public function getVariationAttributes(string $variationAttr, bool $checkIfGroupable): array
```
TheKeymaster, line 93: Please see my other comment regarding to `$checkIfGroupable`. I also don't like abbreviations, please always try to use the full name if it makes sense, so calling the argument simply `$variationAttribute`.

Additionally please typehint the return value with the PHPdoc, as an `array` without a known structure is kinda hard to get into.

src/Wrapper/SeparatedVariation.php
```diff
@@ -79,10 +83,30 @@ public function getVariationGroupKey(): string
...
+    {
+        if (strpos($variationAttr, self::ATTRIBUTE_VALUE_SEPARATOR)) {
```
TheKeymaster, line 95: I think we can use `str_starts_with`, as usually Symfony includes their PHP-x-x polyfills in their dependencies. Otherwise please be explicit to check against `=== 0`, as without this check it doesn't necessarily need to start with the given string, but could be on any position.

src/Wrapper/Product.php
```diff
@@ -210,11 +221,23 @@ protected function processVariations(bool $checkAvailability = true): int
...
+            if (!$defaultImg && $variation->getImage()) {
+                $defaultImg = $variation->getImage();
+            }
+
+            if (!$this->registryService->shouldUseLegacyCallistoUrl() && $variation->getPrice() !== 0.0) {
+                $variationsPriceData[] = [
+                    self::VARIATION_ID => $variation->getId(),
+                    self::PRICE => $variation->getPrice(),
+                    self::IMAGE => $variation->getImage()
+                ];
```
TheKeymaster, line 238: A method should be responsible for getting the proper format.

TheKeymaster: Also why is there an `image` inside of "price data"? Doesn't seem to belong in such a variable.

src/Wrapper/Product.php
```diff
@@ -29,6 +29,12 @@ class Product
 
     public const WRAP_MODE_SEPARATE_VARIATIONS = 1;
 
+    private const VARIATION_ID = 'id';
+
+    private const PRICE = 'price';
+
+    private const IMAGE = 'image';
```
TheKeymaster, line 36: Please move this into a separate class. It seems like this was added here just for convenience, but this should not belong here.

src/Wrapper/Product.php
```diff
@@ -260,6 +283,27 @@ protected function processVariations(bool $checkAvailability = true): int
...
+
+            if (!$hasImage && !empty($cheapestVariationData) && $cheapestVariationData[self::IMAGE]) {
+                $this->item->addImage($cheapestVariationData[self::IMAGE]);
+                $hasImage = true;
+            }
+
+            if (!$hasPrice && $cheapestVariationData[self::PRICE]) {
+                $this->item->addPrice($cheapestVariationData[self::PRICE]);
+                $hasPrice = true;
+            }
```
TheKeymaster, line 300: Put this logic into the class (mentioned in a previous comment, regarding the `const` at the top of this class) responsible for adding images and classes based on the cheapest variation.

src/Wrapper/Product.php
```diff
@@ -260,6 +283,27 @@ protected function processVariations(bool $checkAvailability = true): int
...
+        }
+
+        if (!$hasImage && $defaultImg) {
```
TheKeymaster, line 303: Please don't use abbreviations.
```suggestion
        if (!$hasImage && $defaultImage) {
```

src/Wrapper/Product.php
```diff
@@ -260,6 +283,27 @@ protected function processVariations(bool $checkAvailability = true): int
             $variationsProcessed++;
         }
 
+        if (!empty($variationsPriceData)) {
+            $priceColumn = array_column($variationsPriceData, self::PRICE);
```
TheKeymaster, line 287: This logic here makes sense, but I would have to check every single line of code to grasp what's going on here. Please split it up into multiple smaller functions, so someone reading this understands the code without knowing all the internals.

```php
// Inside a class

/**
 * Adds image and price based on the cheapest variation to the export item.
 *
 * @return int ID of the cheapest variation.
 */
public function addImageAndPrice(Item $item, array $variationsPriceData): int
    $variation = $this->getCheapestVariationData($variationsPriceData);
    $this->addDataToItem($item, $variation);

    return $variation[self::VARIATION_ID];
}
```

tests/RegistryServiceTest.php
```diff
@@ -698,6 +698,41 @@ public function testGetPluginConfigurations(): void
         $this->assertEquals($configData, $this->registryService->getPluginConfigurations());
     }
 
+    /**
+     * @dataProvider shouldUseLegacyCallistUrlTestProvider
+     */
+    public function testShouldUseLegacyCallistUrl(
```
TheKeymaster, line 704: Nice, simple, and easy to read test :)

TheKeymaster (changes requested):

src/Wrapper/SeparatedVariation.php
```diff
@@ -36,6 +37,11 @@ public function isImageAvailable(array $imageAttributeValues, array $variationAt
         foreach ($imageAttributeValues as $imageAttributeValue) {
             foreach ($variationAttributes as $variationAttribute) {
                 $attributeData = explode(self::ATTRIBUTE_VALUE_SEPARATOR, $variationAttribute);
+
+                if (count($attributeData) !== 2) {
```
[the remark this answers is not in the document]
msilinskas: Yes, when product has only one groupable attribute, it's key is the main variation ID. For this reason we can't check the attribute id and value id.

src/Wrapper/SeparatedVariation.php
msilinskas: We are'nt checking if string is starting with value separator. We are checking, if string contains it, if not - we have to get attribute group key. Does this change is still necessary?

src/Wrapper/CheapestVariation.php
```diff
@@ -0,0 +1,83 @@
...
+    public function addVariation(Variation $variation): void
+    {
+        $this->cheapestVariationsData[] = [
+            self::VARIATION_ID => $variation->getId(),
+            self::PRICE => $variation->getPrice(),
+            self::IMAGE => $variation->getImage()
+        ];
+    }
+
+    public function addImageAndPrice(
```
TheKeymaster, line 36: Maybe make it clear in a docblock what it returns.

src/Wrapper/SeparatedVariation.php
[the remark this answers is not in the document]
TheKeymaster: Please add this as a comment.

src/Wrapper/CheapestVariation.php
```diff
@@ -0,0 +1,83 @@
...
+        bool $hasImage = false
+    ): ?int {
+        if (empty($this->cheapestVariationsData)) {
+            $this->setDefaultImage($defaultImage, $hasImage);
+            $this->setDefaultPrice($prices);
+
+            return null;
+        }
+
+        $cheapestVariationData = $this->getCheapestVariation();
```
TheKeymaster, line 48: Here you call it `$cheapestVariationData` but everywhere else you call it `$cheapestVariationsData`. Please make the naming consistent.

src/Wrapper/Product.php
```diff
@@ -194,6 +196,8 @@ protected function processVariations(bool $checkAvailability = true): int
         $baseUnit = null;
         $packageSize = null;
         $variationId = null;
+        $cheapestVariationData = new CheapestVariation($this->item);
```
TheKeymaster, line 199: `Data` suggests unstructured data, but this class isn't really holding any data, its purpose is to map the data from multiple `Variation` entities to an `Item` instance for the export.

src/Wrapper/SeparatedVariation.php
```diff
@@ -85,4 +89,43 @@ public function getVariationGroupKey(): string
...
+    public function getVariationAttributes(string $variationAttributes, int $wrapMode): array
+    {
+        if (strpos($variationAttributes, self::ATTRIBUTE_VALUE_SEPARATOR)) {
+            return explode(self::MULTIPLE_KEYS_SEPARATOR, $variationAttributes);
+        }
+
+        return explode(self::MULTIPLE_KEYS_SEPARATOR, $this->getVariationGroupKeyForImageProcessing($wrapMode));
+    }
+
+    private function isAttributeGroupable(Attribute $attribute): bool
```
TheKeymaster, line 105: Nice helper

TheKeymaster (changes requested):

src/Wrapper/SeparatedVariation.php
```diff
@@ -36,6 +37,15 @@ public function isImageAvailable(array $imageAttributeValues, array $variationAt
         foreach ($imageAttributeValues as $imageAttributeValue) {
             foreach ($variationAttributes as $variationAttribute) {
                 $attributeData = explode(self::ATTRIBUTE_VALUE_SEPARATOR, $variationAttribute);
+
+                /**
+                 * if variation has only one groupable attribute, it's key value will be main variation id.
+                 * For this reason $attributeData will not contain an attribute id and attribute value id.
+                 */
```
TheKeymaster, line 44: Even for comments over multiple lines, always use inline comments.

```suggestion
                // If variation has only one groupable attribute, it's key value will be main variation id.
                // For this reason $attributeData will not contain an attribute id and attribute value id.
```

TheKeymaster (approved):

PR #64, opened 2022-01-21
PLENTY-419 fixing cheapest variation image adding to item

msilinskas:


src/Wrapper/CheapestVariation.php
```diff
@@ -91,4 +91,18 @@ private function setDefaultPrice(array $prices): void
 
         $this->item->addPrice(min($prices));
     }
+
+    /**
+     * @param array<string,string> $cheapestVariationsData
```
TheKeymaster, line 96: Shouldn't this be `array<string, string|float>`, because the array also contains the price of the variation?
```suggestion
     * @param array<string, string|float> $cheapestVariationsData
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #65, opened 2022-01-24
PLENTY-405-decreasing-export-time

dblaichinger:


tests/Wrapper/ProductTest.php
```diff
@@ -974,12 +976,21 @@ private function getProduct(): Product
...
             $this->storeConfigurationMock,
             $this->registryServiceMock,
+            $this->getPropertySelections(),
             $this->itemMock,
             $this->variationEntityMocks,
             Product::WRAP_MODE_DEFAULT
         );
     }
 
+    private function getPropertySelections(): ?PropertySelectionResponse
```
TheKeymaster: In which case would this return `null`? Probably never, otherwise we should be aware of it and this should fail.
```suggestion
    private function getPropertySelections(): PropertySelectionResponse
```

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #66, opened 2022-02-22
plenty-425 exporting external manufacturer name

msilinskas:


src/Wrapper/Product.php
```diff
@@ -362,11 +362,13 @@ protected function addManufacturer(): void
         if (!Utils::isEmpty($manufacturerId)) {
             $manufacturer = $this->registryService->getManufacturer($manufacturerId);
 
-            if (Utils::isEmpty($manufacturer->getName())) {
+            if (Utils::isEmpty($manufacturer->getExternalName()) && Utils::isEmpty($manufacturer->getName())) {
```
TheKeymaster: This could be a simple function like `hasManufacturerNameSet` or something similar.

src/Wrapper/Product.php
```diff
@@ -362,11 +362,13 @@ protected function addManufacturer(): void
...
                 return;
             }
 
-            $vendorAttribute = new Attribute('vendor', [$manufacturer->getName()]);
+            $manufacturerName = ($manufacturer->getExternalName()) ?: $manufacturer->getName();
```
TheKeymaster: I prefer `??` over `?:`. We also may get rid of the braces. I like to add braces for boolean expressions, but in this case I don't really see the benefit :D

```suggestion
            $manufacturerName = $manufacturer->getExternalName() ?? $manufacturer->getName();
```

tests/Wrapper/ProductTest.php
```diff
@@ -1031,6 +1063,20 @@ public function cheapestVariationIsExportedTestProvider(): array
...
     }
 
+    public function correctManufacturerIsExportedTestProvider(): array
+    {
+        return [
+            'manufacturer has external name, external name is exported' => [
+                'ManufacturerResponse/one.json',
+                'cat=Sessel+%26+Hocker&cat_url=%2Fwohnzimmer%2Fsessel-hocker%2F&vendor=externalNameA',
+            ],
+            'manufacturer has not external name, original name is exported' => [
```
TheKeymaster:
```suggestion
            'manufacturer has no external name, original name is exported' => [
```

TheKeymaster (changes requested):

src/Wrapper/Product.php
msilinskas: ?: checks for an empty strings but ?? - not. We are getting an empty string if the external name is not provided for manufacturer:
image

With ?? we are getting the following error:
Something went wrong. Message:                                         
         FINDOLOGIC\Export\Exceptions\EmptyValueNotAllowedException: Empty      
         values are not allowed for "attribute" values

TheKeymaster: Wow.. typical Plenty :face_exhaling:

TheKeymaster (approved):

PR #67, opened 2022-03-08
PLENTY-431 Export variant id of cheapest variant

TobiasGraml11:


tests/MockData/Pim/Variations/response_with_no_main_variation.json
```diff
@@ -2829,11 +2829,11 @@
       "markets": null,
       "salesPrices": [
         {
-          "salesPriceId": 1,
+          "salesPriceId": 0,
           "price": 79
         },
         {
-          "salesPriceId": 2,
+          "salesPriceId": 1,
```
dblaichinger: Why was this necessary? Also using a test response without main variation seems not the right case here, how about using `response_for_lowest_price_test.json` instead?

src/Wrapper/Product.php
```diff
@@ -318,9 +318,9 @@ protected function processVariations(bool $checkAvailability = true): int
             $this->item->addProperty($packageSizeProperty);
         }
 
-        if ($variationId) {
+        if ($this->cheapestVariationId || $variationId) {
             $variationIdProperty = new Property('variation_id');
-            $variationIdProperty->addValue((string)$variationId);
+            $variationIdProperty->addValue((string)($this->cheapestVariationId ?? $variationId));
```
dblaichinger: Extracting the value in an own line would help improve readability as the amount of brackets is quite high here.

dblaichinger (changes requested):

tests/MockData/Pim/Variations/response_with_no_main_variation.json
TobiasGraml11: This was needed, to set the second variant as cheapest. Otherwise the test would be irrelevant.

I used this data, as it was already used by the previous test, but I switched it now.

TobiasGraml11: This was a leftover from a different test adaption without removing the main variant id implementation

dblaichinger (approved):

PR #74, opened 2022-10-10
PLENTY-457 Use account as data source

TobiasGraml11:


tests/Command/GenerateTokenCommandTest.php
```diff
@@ -49,9 +49,7 @@ public function testTokenIsGenerated(): void
         $expectedToken = 'access token';
 
         $commandTester = new CommandTester($this->command);
-        $commandTester->execute([
-            'shopkey' => 'ABCDABCDABCDABCDABCDABCDABCDABCD'
-        ]);
+        $commandTester->execute([]);
```
TobiasGraml11: There should be another fix for it. The command should still work with a shopkey as argument.

PR #75, opened 2023-01-19
PLENTY-461 Export product dimensions

TobiasGraml11:


tests/ConfigTest.php
```diff
@@ -71,8 +80,16 @@ public function testDataCanBeFetchedFromCustomerLoginData(): void
                     'availability_id' => $expectedAvailabilityId,
                     'price_id' => $expectedPriceId,
                     'rrp_id' => $expectedRrpId,
-                    'exporter' => 'REST',
```
TobiasGraml11, line 74: The `exporter` config was removed within the same story the new configurations were added.

tests/MockData/Pim/Variations/response.json
```diff
@@ -26511,8 +26511,8 @@
         "availableUntil": null,
         "releasedAt": null,
         "name": "Coole Variante",
-        "weightG": 17,
-        "weightNetG": 17,
+        "weightG": 0,
+        "weightNetG": 0,
```
TobiasGraml11, line 26515: No need for that to be set, as this is tested within other tests anyway

aserstobitov (approved):

PR #82, opened 2023-05-12
PLENTY-471 upgrade the exporter to libflexport v3 (PLENTY-475)

Mariusslo:


.vscode/launch.json
TobiasGraml11, line 1: Please remove the `.vscode` folder and add it to the `.gitignore` file

src/Request/ItemRequest.php
```diff
@@ -13,7 +13,7 @@ public function __construct(
         ?string $lang = null,
         ?string $name = null,
         ?string $manufacturerId = null,
-        ?int $id = null,
+        ?int $id = 107,
```
TobiasGraml11:
```suggestion
        ?int $id = null,
```

src/Exporter/CsvExporter.php
TobiasGraml11, line 1: We can remove the `CsvExporter` and it's helper classes (if some exist). We won't add support for it with libflexport v3 anyway.

src/Command/ExportCommand.php
```diff
@@ -155,8 +155,18 @@ private function shouldStartExportIfFileAlreadyExists(
         }
 
         $exportFileLocation = Utils::env('EXPORT_DIR', Exporter::DEFAULT_LOCATION);
-        $isCsvExporter = $exporterType === Exporter::TYPE_CSV;
-        if (!$isCsvExporter || !file_exists($exportFileLocation . '/findologic.csv')) {
+        $fileExists = false;
+        switch ($exporterType) {
```
TobiasGraml11: Let's remove the command option entirely and write the code like only XML exists.

phpunit.xml.dist
```diff
@@ -16,4 +16,8 @@
       <directory>./tests</directory>
     </testsuite>
   </testsuites>
-</phpunit>
+
+  <extensions>
+    <extension class="FINDOLOGIC\PlentyMarketsRestExporter\Tests\Helper\BypassFinalHook" />
+  </extensions>
+</phpunit>
```
TobiasGraml11: Please add a newline to the end of the file

src/Wrapper/Product.php
```diff
@@ -276,6 +277,10 @@ protected function processVariations(bool $checkAvailability = true): int
                 $this->item->addMergedAttribute($attribute);
             }
 
+            foreach ($variation->getImages() as $image) {
+                $this->item->addImage($image);
+            }
```
TobiasGraml11: There is already an image added at line 233, we should probably combine those.

src/Wrapper/Product.php
```diff
@@ -297,13 +302,8 @@ protected function processVariations(bool $checkAvailability = true): int
...
         }
 
-        // VatRate should be set from the last variation, therefore this code outside the foreach loop
-        if (isset($variation) && $variation->getVatRate() !== null) {
-            $this->item->setTaxRate($variation->getVatRate());
-        }
-
         if ($insteadPrices) {
-            $this->item->setInsteadPrice(min($insteadPrices));
+            $this->item->setOverriddenPrice(new OverriddenPrice());
```
TobiasGraml11: 1. There is no value set here
2. Please rename all occurrences of `insteadPrice` to `overriddenPrice`

tests/Command/ExportCommandTest.php
```diff
@@ -70,15 +70,19 @@ public function testCommandFailsWhenUsingWrongFormattedShopkey(): void
         ]);
     }
 
-    public function testExportDoesNotStartWhenFileAlreadyExists(): void
+    /**
+     * @dataProvider exportTypesProvider
```
TobiasGraml11: After getting rid of the CSV part, we can change everything to XML-only here as well

tests/Exporter/ExporterTest.php
```diff
@@ -208,7 +209,7 @@ public function testExceptionIsCaughtForProdEnvironment(int $type): void
     public function testFileNameIsChangedWhenSet(): void
     {
         $this->fileNamePrefix = 'findologic.new.funny';
-        $this->fileExporterMock = LibFlexportExporter::create(LibFlexportExporter::TYPE_CSV);
+        $this->fileExporterMock = LibFlexportExporter::create(ExporterType::CSV);
```
TobiasGraml11: CSV is still used here

tests/Helper/ItemHelper.php
TobiasGraml11, line 1: Great helper class :+1:

tests/Wrapper/ProductTest.php
```diff
@@ -537,7 +536,7 @@ public function testPriceAndInsteadPriceIsSetByLowestValues(): void
         $item = $product->processProductData();
 
         $this->assertEquals(['' => 50], $item->getPrice()->getValues());
-        $this->assertEquals(100, $item->getInsteadPrice());
+        // $this->assertEquals(100, $item->getInsteadPrice()); //Temporarily remove because the fn doesnt exist anymore
```
TobiasGraml11: Please change the test function to test the `overriddenPrice`

tests/Wrapper/ProductTest.php
```diff
@@ -719,15 +716,19 @@ public function testFreeTextFieldsAreNotExportedAccordingToConfiguration(
...
+            $this->assertEquals($expectedAttributeValue, $attributesMap[$key][0]);
+        }
     }
 
     public function testTooLongFreeTextFieldsAreIgnored(): void
     {
-        $expectedValue = 'cat=Sessel+%26+Hocker&cat_url=%2Fwohnzimmer%2Fsessel-hocker%2F&free1=0000000000';
+        $expectedAttributeValues = [
+            'cat' => 'Sessel & Hocker',
+            'cat_url' => '/wohnzimmer/sessel-hocker/', 'free1' => '0000000000'
```
TobiasGraml11:
```suggestion
            'cat_url' => '/wohnzimmer/sessel-hocker/',
            'free1' => '0000000000'
```

TobiasGraml11 (changes requested):

rimvydas-hashtages:
@TobiasGraml11 you can now review the changes

src/Command/ExportCommand.php
```diff
@@ -155,8 +155,10 @@ private function shouldStartExportIfFileAlreadyExists(
         }
 
         $exportFileLocation = Utils::env('EXPORT_DIR', Exporter::DEFAULT_LOCATION);
-        $isCsvExporter = $exporterType === Exporter::TYPE_CSV;
-        if (!$isCsvExporter || !file_exists($exportFileLocation . '/findologic.csv')) {
+
```
TobiasGraml11, line 144: This function still has an `$exporterType`argument, we can remove that as well.

src/Command/ExportCommand.php
```diff
@@ -80,7 +80,7 @@ protected function configure()
                 Exporter::TYPE_CSV,
                 Exporter::TYPE_XML
             ),
-            Exporter::TYPE_CSV
+            Exporter::TYPE_XML
```
TobiasGraml11: We can remove this whole `addOption` block.

src/Wrapper/Product.php
```diff
@@ -229,7 +230,10 @@ protected function processVariations(bool $checkAvailability = true): int
 
             $useCallistoUrl = $this->registryService->getPlentyShop()->shouldUseLegacyCallistoUrl();
             if (!$itemHasImage && $variation->getImage() && $useCallistoUrl) {
-                $this->item->addImage($variation->getImage());
+                $this->item->setAllImages($variation->getImages());
+                $itemHasImage = true;
+            } elseif (!$itemHasImage) {
+                $this->item->setAllImages($variation->getImages());
```
TobiasGraml11: Let's remove the `elseif` block and keep the logic as it is right now.
Right now, it is handled within `$cheapestVariations->addImageAndPrice`, instead of an additional `elseif`.

src/Wrapper/Product.php
```diff
@@ -297,13 +301,10 @@ protected function processVariations(bool $checkAvailability = true): int
...
 
-        // VatRate should be set from the last variation, therefore this code outside the foreach loop
-        if (isset($variation) && $variation->getVatRate() !== null) {
-            $this->item->setTaxRate($variation->getVatRate());
-        }
-
         if ($insteadPrices) {
-            $this->item->setInsteadPrice(min($insteadPrices));
+            $overriddenPrice = new OverriddenPrice();
+            $overriddenPrice->setValue(min($insteadPrices));
```
TobiasGraml11: Please rename `$insteadPrices` to `$overriddenPrices`

.env
```diff
@@ -22,4 +22,4 @@ EXPORT_WEIGHT_UNIT=g
 DEBUG=true
 IMPORT_DATA_URL=https://account.findologic.com/api/v1/shops/%s/importData
 
-ACCOUNT_PAT=""
+ACCOUNT_PAT="xRPBSki22bzO75sOfehYjN0m12BBZia7RwfKyURS"
```
TobiasGraml11: We should not commit any PAT, especially not within public repositories.

Please remove this and remove/regenerate the token from your account.

TobiasGraml11 (changes requested):

.env
rimvydas-hashtages: Can I rename this to .env.example and add .env to gitignore? Because thats usually how I do to avoid this

TobiasGraml11: Sure, let's do it like that

src/Wrapper/Product.php
rimvydas-hashtages: without the elseif block it wouldnt pass because it checks $useCallistoUrl there

TobiasGraml11: Yeah, that's how it was done before as well. I don't want to change too much logic within this story.
If the check does not pass, it should just add the default image at a later stage, like it is done right now.

rimvydas-hashtages: Then data provided in the tests must be wrong cause it keeps failing the tests
image

TobiasGraml11: The expected value was changed within the data provider. We will need to revert that change.

rimvydas-hashtages: Okay changed it. You can review the PR again

src/Wrapper/Product.php
```diff
@@ -277,7 +278,7 @@ protected function processVariations(bool $checkAvailability = true): int
             }
 
             $prices[] = $variation->getPrice();
-            $insteadPrices[] = $variation->getInsteadPrice();
+            $overriddenPrices[] = $variation->getInsteadPrice();
```
TobiasGraml11: This function still includes `insteadPrice`

tests/Command/ExportCommandTest.php
TobiasGraml11: We can get rid of this data provider now, as it only has one value anyway

tests/Exporter/ExporterTest.php
```diff
@@ -98,12 +98,7 @@ protected function tearDown(): void
     public function exporterTypeProvider(): array
     {
         return [
-            'Exporter type is CSV' => [
```
TobiasGraml11, line 101: Please completely remove the data provider

TobiasGraml11 (changes requested):

src/Wrapper/Product.php
rimvydas-hashtages: ?

TobiasGraml11 (approved):

PR #85, opened 2023-09-07
FINDO-9977 Support new version within importer

TobiasGraml11:


src/Wrapper/Product.php
```diff
@@ -212,10 +212,10 @@ protected function setTexts(): void
             }
 
             if (trim($text->getShortDescription()) !== '') {
-                $this->item->addSummary($text->getShortDescription());
+                $this->item->addSummary(preg_replace('/[\x00-\x1F\x7F]/u', '', $text->getShortDescription()));
```
TobiasGraml11, line 215: This bug was found, when testing the export of 2DC7C0FA293A3EE65E97095B31B5F962.

I am reusing the fix we already needed to add within Shopware 6.

rimvydas-hashtages: Maybe it makes sense to escape html chars also so that it wouldn't break the xml structure?

TobiasGraml11: Everything is wrapped within a [CDATA](https://www.tutorialspoint.com/xml/xml_cdata_sections.htm) tag, HTML is supported within those.

XML just doesn't support this weird control characters:
image

rimvydas-hashtages (approved):

PR #88, opened 2023-09-11
PLENTY-481 Add fixes found after deployment

TobiasGraml11:


src/Wrapper/Product.php
```diff
@@ -694,4 +706,24 @@ private function addOrdernumber(string $ordernumber): void
 
         $this->item->addOrdernumber(new Ordernumber($ordernumber));
     }
+
+    private function isMissingRequiredFields(): ?string
```
rimvydas-hashtages: Not sure how the error tracking works, but maybe it would make more user friendly to list all the missing fields

rimvydas-hashtages (approved):
Code looks good to me

PR #91, opened 2023-11-09
Plenty-483-imports-fail-due-to-insufficient-memory

aserstobitov:


src/Utils.php
```diff
@@ -49,6 +49,8 @@ public static function sendIterableRequest(Client $client, Request $request): ar
 
         $responses = [];
         $lastPage = false;
+        $request->setItemsPerPage($client->getConfig()->getItemsPerPage());
```
TobiasGraml11: I would say the cleaner solution is to provide the config as a parameter. Every class calling this function should already have the config loaded.

src/Exporter/Exporter.php
```diff
@@ -217,7 +217,7 @@ protected function exportProducts(): void
             $products = $this->getItems($page);
             $variations = $products->getAllIds() ? $this->getItemVariations($products->getAllIds()) : [];
             $this->wrapData(count($products->all()), $products, $variations, $propertySelection);
-            $this->offset += ItemVariationRequest::$ITEMS_PER_PAGE;
+            $this->offset += $this->config->getItemsPerPage();
```
TobiasGraml11: Guess this one will be safer:
```suggestion
            $this->offset += $this->itemRequest->getItemsPerPage();
```

src/Client.php
```diff
@@ -236,6 +236,11 @@ public function getAccessToken(): string
         return $this->accessToken;
     }
 
+    public function getConfig(): Config
```
TobiasGraml11: This can be removed, after the config is provided to the utils function via parameter

src/Utils.php
TobiasGraml11: But the placement ist great here, thanks for coming up with this!

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #94, opened 2024-03-05
Plenty-485-properties-consider-the-client

aserstobitov:


src/RegistryService.php
```diff
@@ -707,22 +707,35 @@ public function createPlentyShop(): void
...
+                if ((int)$option->getValue() === $storeIdentifier) {
+                    $isExportableByClient = true;
+                }
             }
         }
 
-        return false;
+        if (!$isClientAssigned) {
+            $isExportableByClient = true;
+        }
```
TobiasGraml11, line 739: Why do we need to make it exportable by client, when no client is assigned? Is this how Plenty handles the data as well?

TobiasGraml11 (approved):

PR #95, opened 2024-05-22
PLENTY-489-retry-requests-with-exceptions

aserstobitov:


src/Client.php
```diff
@@ -90,10 +92,28 @@ public function send(Request $request): ResponseInterface
     {
         $this->handleRateLimit();
         $this->handleLogin();
+        
+        $endpoint = $request->getUri()->__toString();
+        $request = $request->withUri($this->buildRequestUri($endpoint));
+        try {
```
TobiasGraml11: It's probably best to wrap this inside a do...while instead of recalling the function again with an already adapted request class.

Something like:
```php
do {
    try {
        $response = $this->sendRequest($request, $request->getParams());
        $this->handleResponse($request, $response);

        return $response;
    } catch (...) {
        // log error

        // increment

        // sleep
    }
} while (!$request->isRetryLimitReached());

throw new Exception('Maximum retry limit reached without success');
```

src/Request/Request.php
```diff
@@ -28,4 +29,23 @@ public function getParams(): array
...
+        $this->retryCounter++;
+    }
+
+    public function isRetryLimitReached(): bool
+    {
+        if ($this->retryCounter == 5) {
+            return true;
+        }
+
+        return false;
```
TobiasGraml11:
```suggestion
        return $this->retryCounter >= self::MAX_RETRIES;
```

tests/ClientTest.php
```diff
@@ -95,6 +95,7 @@ public function badResponseProvider(): array
                 'response' => new GuzzleResponse(401, [], 'OOF you are not logged in bro...'),
                 'expectedException' => AuthorizationException::class,
                 'expectedExceptionMessage' => 'The REST client is not logged in.',
+                'expectedCallTimes' => 2
```
TobiasGraml11, line 99: Shouldn't it only be called once, if the exception is non-retryable?

TobiasGraml11 (changes requested):

tests/ClientTest.php
aserstobitov: @TobiasGraml11 It's called 2 times because the 1st time it needs to login, and 2nd time is the actual request

TobiasGraml11: Ok, makes sense 👍

src/Request/Request.php
```diff
@@ -28,4 +31,19 @@ public function getParams(): array
...
+    }
+
+    public function incrementRetryCounter(): void
+    {
+        $this->retryCounter++;
+    }
+
+    public function isRetryLimitReached(): bool
+    {
+        return $this->retryCounter == self::RETRY_LIMIT;
```
TobiasGraml11: We need to make sure to also check the higher counts, in case the increment function is even slightly misused in future changes.

```suggestion
        return $this->retryCounter >= self::RETRY_LIMIT;
```

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

